PR #19, opened 2017-04-27
Rename Binary To Decimal.c to Conversions/Binary To Decimal.c

theycallmemac:


dynamitechetan:
Can we have file name without spaces please?

PR #27, opened 2017-07-04
Add 2 more algorithms (Fibonacci and Factorial)

Max285:
I saw that in the java version there was the factorial algorithm, so, i thought that it would be nice to have the same but this time made with C. And I also made a Fibonacci algorithm that asks for a number and than it tells which number is in the N position according to the Fibonacci sequence. Hope you like it :)

### DIFF
--- a/fat.c
+++ b/fat.c
@@ -1,0 +1,20 @@
+#include <stdio.h>
+
+int fat(int number){
+	if (number == 1) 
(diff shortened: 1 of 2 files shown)


0oo0:
Thanks for the PR 🎉. Looking forward to having more contributions from you.

Issue #42, opened 2017-09-29
Travis CI


AnupKumarPanwar:
Add a .travis.yml file to this repository.
Refer to this doc https://docs.travis-ci.com/user/languages/c/

ankitrgadiya:
For this first we need makefiles. I'm working on it

PR #48, opened 2017-09-29
Global and Local Makefiles added

ankitrgadiya:
Makefiles added in all directories in order to compile with ease using Travis CI

Few source file names have space in them and so I explicitly specified them in Makefile.
#42


christianbender:
@ankitrgadiya  Makefiles are not needed. In this repo are many projects that contains only functions. In that cases Makefiles are not useful.

PR #54, opened 2017-10-01
added tree problem 'common nodes in two BSTs'

nkkumawat:
Find common nodes into two BSTs


christianbender:
@nkkumawat  This is a C repository. That means none C++ code. ```#include<stack>``` is a data structure of the C++ standard library.

Issue #37, opened 2017-09-28
Add more algorithms


AnupKumarPanwar:
Please add more algorithms to this repo. 

See the structure and README of this repo for reference
https://github.com/TheAlgorithms/Python

prakamya-mishra:
I would like to contribute merge sort algorithm for hactoberfest. when should i give a pull request for   it after 1st october ?

prakamya-mishra:
Sorry I meant code for All types of Linked List not merge sort.

AnupKumarPanwar:
Yeah sure, go ahead and send pull request

prakamya-mishra:
Please look over #38

prakamya-mishra:
rest all i will send after october 1st

AnupKumarPanwar:
Merged

prakamya-mishra:
Thank you

prakamya-mishra:
I have a Question:
All my pull request should reference to this issue then only will it get counted in hacktoberfest?

kartikey-singh:
Please review the new fat.c added by me (Dynamic Programming Version)

jebinphilipose:
Please review my PR's (#51 and #52)

vagita24:
hi @AnupKumarPanwar 
I am a begineer. I want to contribute in Algos. Can you tell me the name of algos to contribute and language to use.

prakamya-mishra:
Please review my PR's #53

MadhavBahlMD:
Hey.
Operating systems related algorithms are missing.
If you want, i can add some?

hegdenaveen1:
Is the Project Euler Branch in this repo serious?

christianbender:
@hegdenaveen1  of course

Rupeshiya:
hi
i have added sorting of linked list using selection sort and also stack implementation by using array 
look after #115

maysanchez:
Hi @AnupKumarPanwar  
Could you review my PR #138, please ?

JRaiden16:
> 
> 
> Please add more algorithms to this repo.
> 
> See the structure and README of this repo for reference
> https://github.com/TheAlgorithms/Python

Did you want some algorithms in particular? I have one that searches the GCD of two numbers by using Euclid's Algorithm, and another one that uses that GCD to code a message using RSA Algorithm

deadPix3l:
[...]

PR #59, opened 2017-10-02
Add Sieve of Eratosthenes

danielgrimshaw:
This commit adds a sieve of Eratosthenes algorithm.

This is for issue #37


christianbender:
Move your file **primes.c** in the directory **Misc**

PR #93, opened 2017-10-30
To reverse a singly linked list.

kaushal10:


christianbender:
@kaushal10  Can you write a test scenario in main?

PR #94, opened 2017-10-31
Create cesar cipher

brumazzi:


christianbender:
@brumazzi 

Can you remove **malloc.h** and associated functions?
Because malloc.h is quite Linux specific and additional deprecated.
The code in this repo should be platform independent.

PR #102, opened 2017-11-11
add heapsort to sorts folder

koseokkyu:
I added heapsort.c file, jumpsearch file to sorts folder and search folder. 

There seems to be no difference except to enter the size of the array when compared to a Python file.

Please check and pull request. I want your feedback.

### DIFF
--- a/Sorts/HeapSort.c
+++ b/Sorts/HeapSort.c
@@ -1,0 +1,67 @@
+#include <stdio.h>
+
+void heapify(int *unsorted, int index, int heap_size);
+void heap_sort(int *unsorted, int n);
+
+int main() {
+	int n = 0;
+	int i = 0;
+	char s[50];
+	char oper;
+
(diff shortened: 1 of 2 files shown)


Sorts/HeapSort.c
```diff
@@ -0,0 +1,67 @@
...
+	int n = 0;
+	int i = 0;
+	char s[50];
+	char oper;
+
+	int* unsorted;
+	printf("Enter the size of the array you want\n");
+	scanf("%d", &n);
+	unsorted = (int*)malloc(sizeof(int) * n);
+	while (getchar() != '\n');
```
AnupKumarPanwar: Keep it simple! Input integers directly rather than using atoi for casting.

koseokkyu: Are you saying that you want to initialize the array and implement it?

PR #109, opened 2017-12-17
New Algorithm - Repeating_1's_Square

ghodakegitanjali:
Added new algorithm to MISC. Fix for issue "Add more algorithms".

This program accepts user input of number of consecutive/repeating 1's. And then calculates it's square.
E.g. If user enters number of consecutive 1's = 5 then it calculates (11111)^2 = 123454321.
User may enter any number of consecutive 1's  to be squared (less than or equal to 2^32 - 1). 
This program uses innovative logic & pattern recognition to calculate the result within fraction of seconds. 
Without this approach high end computers may also take few hours to calculate it. The result is calculated in three parts ' front' 'middle' & 'back'.
Similarly in future this can be extended with a logic to calculate the square of other consecutive/repeating numbers.


christianbender:
What for benefit has your algorithm?

PR #122, opened 2018-03-18
Add three new codes 

ashishtheman:
Three new codes:
1.Catalan numbers.
2.Factorial_trailing_zeroes
3.Collatz conjecture


christianbender:
@ashishtheman I will look in your code tomorrow and sent your feedback.

christianbender:
@ashishtheman  

**to Collatz:** some suggestions
- Notice you can remove the line ```printf("%d->",curr_no);``` from each if-branch an sets the line on the end of the while. 
- Write a small user interface what the program does and was kind of data the program want.

PR #132, opened 2018-03-23
Create Catalan.c

anujms1999:
Program for printing Catalan numbers. Many problems can be solved easily if we know Catalan numbers and above program can be used to find Catalan


vaithak:
Catalan numbers code is already there in the misc folder.

PR #135, opened 2018-06-28
Create iterativeQS.c

akshaykumar123:
iterative sorting algorithm


sorting/iterativeQS.c
```diff
@@ -0,0 +1,35 @@
+int partition(void **A, int n, void *pivot){
```
christianbender: Please ad a ```main``` function and a test case for testing.

PR #138, opened 2018-08-13
Anagrams

maysanchez:
Hey ! 
I added a program that evaluates if two words are anagrams.

### DIFF
--- a/misc/areAnagrams.c
+++ b/misc/areAnagrams.c
@@ -1,4 +1,5 @@
 #include <stdio.h>
+#include <string.h>
  
  //Function prototype
 int check_anagram(char [], char []);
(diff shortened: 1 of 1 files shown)


misc/areAnagrams.c
```diff
@@ -0,0 +1,53 @@
...
+int check_anagram(char [], char []);
+ 
+
+ //main
+int main()
+{
+  char a[100], b[100];
+ 
+  printf("Enter a string\n");
+  gets(a);
```
ashwek: [Why is the gets function so dangerous that it should not be used? - StackOveflow](https://stackoverflow.com/questions/1694036/why-is-the-gets-function-so-dangerous-that-it-should-not-be-used)

PR #141, opened 2018-09-27
I added decimal_to_any_base and any_base_to_any_base

angelos97:


ashwek:
@angelos97 you must declare functions before you use them.

PR #147, opened 2018-10-02
You shall not character my base!

JRaiden16:
What if I try to put a character when the program read the base? The while cycle added prevents that


ashwek:
@JRaiden16 please resolve conflicts

JRaiden16:
I don't have write access :(

PR #152, opened 2018-10-03
Binary Search Updated

SagnikDev:


ashwek:
Your changes are not ideal. You should split the code into different parts instead of having everything in the main function (the way it was earlier - search algorithm in a different function & driver code in the main)
& you removed all the comments, which is the opposite of what you should be doing.

PR #159, opened 2018-10-03
Added base64 and Programmer humor sort functions

deadPix3l:
#37 hacktoberfest

added some abysmal sorting functions. I don't recommend ever using them, but they are good fun.


deadPix3l:
Added base64

ashwek:
@deadPix3l thank you for contributing base64. Please move the file to **conversions** directory.
Programs in **jokes** directory are not useful to anyone in any way, so please remove them from the PR.

PR #160, opened 2018-10-04
removed conio.h

jamesjallorina:
removed conio.h as it is deprecated header file


ashwek:
Thank you for contributing but similar changes were accepted from an earlier PR #146

PR #164, opened 2018-10-08
Removed duplicate occurences

aparimeet:


ashwek:
Thank you for contributing but there is no duplicate **HeapSort.c** & BucketSort.c should remain in **sorting** & not **Sorts**

PR #166, opened 2018-10-08
Q6 project Euler

adityasheth2000:
Please check this method of solving question in O(1) time complexity


ashwek:
@adityasheth2000 please move your code to **project_euler/Problem 06/** directory

PR #168, opened 2018-10-08
Application of Merge sort

adityasheth305:
This merge sort takes two arrays. Then it sorts both the arrays with one merge sort in ascending order and the other array's element are changed corresponding to the elements in first array.


ashwek:
please add **.c** file extension

PR #169, opened 2018-10-08
Please check the solution for Q5 of Project Euler

adityasheth305:


Project Euler/Q5 Project Euler
```diff
@@ -0,0 +1,24 @@
+#include<stdio.h>
+int main()
+{
+        int n;
+        scanf("%d",&n);
```
buffet: Maybe add a prompt for the user to know what to enter (problem states 20)

Project Euler/Q5 Project Euler
```diff
@@ -0,0 +1,24 @@
...
+                for(int j=1;j<=n;j++)
+                {
+                //      int flag=1;
+                        if(i%j!=0)
+                        {
+                                flag=0;
+                                break;
+                        }
+                }
+                if(flag==1)
```
buffet: Just a style thing but comparison to 1 is something many people don't want to see.

`if (flag)` is equivalent. (You're asking "is it true, that this is true?" instead of "is this true?")

Project Euler/Q5 Project Euler
```diff
@@ -0,0 +1,24 @@
...
+                {
+                        printf("%d",i);
```
buffet: Missing linebreak at the end

PR #185, opened 2018-10-18
asking user for searching again

shawabhishek:


searching/LinearSearch.c
```diff
@@ -10,7 +10,8 @@ int linearsearch(int *arr, int size, int val){
 }
 
 void main(){
-	int s,i,v;
+	while(True){
```
ashwek: it should be _**t**rue_ not _**T**rue_

searching/LinearSearch.c
```diff
@@ -24,4 +25,9 @@ void main(){
 		printf("Value %d is in the array.\n", v);
 	else
 		printf("Value %d is not in the array.\n", v);
+	printf("Want to search again press 1");
+	scanf("%d",y);
```
ashwek: it should be **scanf("%d", &y);**

searching/LinearSearch.c
```diff
@@ -24,4 +25,9 @@ void main(){
...
+	if(y!=1)
+	exit;
```
ashwek: **exit(0)** is a function. You missed the parenthesis.
You need to _include_ **stdlib.h** for __exit(0)__ to work.
or you could simply use **break** instead of exit

PR #200, opened 2018-10-27
An efficient solution.

akash19jain:


project_euler/Problem 01/sol1.c
```diff
@@ -1,19 +1,24 @@
...
-	int sum = 0;
-	scanf("%d", &n);
-	for (int a = 0; a < n; a++) {
-		if ((a % 3 == 0) || (a % 5 == 0)) {
-			sum += a;
-		}
-	}
+int main(){
+    int t; 
+    scanf("%d",&t);
```
christianbender: This line is awkward. Can you write a ```while``` condition such that the user can exit the program bei typing -1 ?
Otherwise you can remove the full while-loop, because the problem definition is with your efficient solution solve.

akash19jain: The question mentioned that the First Line contains an Integer T which denotes the Number of test cases. Hence the above lines.

christianbender: You can put in while condition for breaking. For example if the user type ```-1``` the program terminates.

PR #201, opened 2018-10-27
A slight change in stack program

abdulraouf9567:


data_structures/stack/main.c
```diff
@@ -7,7 +7,7 @@ void peep();
 void update();
 int main()
 {
-    int n,a[100],top=0;
+    int n,a[100],top=-1;
     //function for pushing the element 
     void push()
```
christianbender: Can you remove the nested functions such as:
* push(...)
* pop()
* peep(...)  
Because nested functions are not a standard in C

data_structures/stack/main.c
```diff
@@ -19,15 +19,16 @@ int main()
...
         {
             printf("\nstack is empty");
         }
         else
         {
             int item;
-            item=a[top];
-            top=top-1;
+            
+            a[top]=item;
```
christianbender: This is a logical error. Because you do not assign a value to the variable ```item```. After that you print it...

data_structures/stack/main.c
```diff
@@ -19,15 +19,16 @@ int main()
     //function for poping the element out
     void pop()
```
christianbender: Can you remove the nested functions like:
* push
* pop
* peep
* update
Because this isn't a standard in C

PR #202, opened 2018-10-28
Added 3n+1

gerroo:
Added the famous 3n+1 algorithm.


christianbender:
@gerroo We already have the collatz-algorithm in our repo. But you can refactor the existing [version](https://github.com/TheAlgorithms/C/blob/master/misc/Collatz.c)

PR #203, opened 2018-10-28
copy

nickassassin:


christianbender:
@nickassassin Sorry but we already have a linear search in this repo. But you can refactoring the existing [version](https://github.com/TheAlgorithms/C/blob/master/searching/LinearSearch.c)

PR #213, opened 2018-12-08
Create perfectNumber.c

TarunRaoK98:
This is a simple algorithm to check if a number is a 'perfect number' or not.

### DIFF
--- a/misc/perfectNumber.c
+++ b/misc/perfectNumber.c
@@ -1,0 +1,21 @@
+#include <stdio.h>
+
+int isPerfectNumber(int n)
+{
+  int sum = 0;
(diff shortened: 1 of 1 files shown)


misc/perfectNumber.c
```diff
@@ -0,0 +1,21 @@
...
+{
+  int n = 0;
+  printf("Number : ");
+  scanf("%d", &n);
+	
+	if(isPerfect(n))
+		printf("%d is a Perfect Number\n", n);
+	else
+		printf("%d is not a Perfect Number\n", n);
+}
```
christianbender: Please add ```return 0;``` in your main-function. Or use a makro like ```exit(0)```.

PR #222, opened 2019-02-19
BugFix : loop out of range will cause stack smashing

54shady:
Data Sorted
    Shell Sort Big O Notation:
    --> Best Case: O(n log(n))
    --> Average Case: depends on gap sequence
    --> Worst Case: O(n)

    Time spent sorting: 0.000006
    *** stack smashing detected ***: <unknown> terminated
    Aborted


yanglbme:
Thanks @54shady

PR #224, opened 2019-03-21
changed algorhytm to create binary number

Aurvandill137:
instead of using an array to save the binary value we write it immeaditly
also added the possibility to enter negative umbers

### DIFF
--- a/conversions/decimal _to_binary.c
+++ b/conversions/decimal _to_binary.c
@@ -1,7 +1,27 @@
 #include <stdio.h>
 #include <stdlib.h>
 
-#define MAXBITS 100
+void mask(int number){
+
+	//create bitmask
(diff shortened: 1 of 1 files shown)


ashwek:
@Aurvandill137 
If you input _5_ the output is _00000000000000000000000000000101_ (32 bits - assuming ``int`` takes 4 bytes).
I think the leading _0's_ are unnecessary and should be avoided.

Aurvandill137:
its just a representation of how the number is stored in memory - since an integer is (under windows)4 byte all 32 bits will get represented 
maybe to a later standpoint i'll implement this functionality

Aurvandill137:
now the programm only writes zeors after the first one is encountered

PR #233, opened 2019-06-05
Changed scanf to fgets

VilopusDoom:
Scanf is generally considered to be less secure than fgets. If you need bigger than 100 chars, change the array length, and the fgets statement accordingly

### DIFF
--- a/conversions/binary_to_decimal.c
+++ b/conversions/binary_to_decimal.c
@@ -1,25 +1,35 @@
 /**
  * Modified 07/12/2017, Kyler Smith
- * 
+ * Modified 06/05/2019, Pocholo Arangote
+ *
  */
 
 #include <stdio.h>
-
+#include <stdlib.h>
+#include<ctype.h>
 int main() {
 
-	int remainder, number = 0, decimal_number = 0, temp = 1;
-	printf("/n Enter any binary number= ");
-	scanf("%d", &number);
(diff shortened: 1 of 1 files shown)


conversions/binary_to_decimal.c
```diff
@@ -1,25 +1,35 @@
...
-		temp = temp*2;		//used as power of 2
-	
-	}
-
-	printf("%d\n", decimal_number);
+    int remainder, number = 0, decimal_number = 0, temp = 1;
+    
+    char binary_number[100] = {};
+    char* binary_number_ptr;
+    printf("/n Enter any binary number= ");
```
ashwek: it should be ``\n`` not ``/n``

ashwek:
@VilopusDoom thank you for contributing.
If the input is ``102`` the output is ``6``. It would be great if you would check if the input is a binary number or not.

conversions/binary_to_decimal.c
```diff
@@ -1,25 +1,35 @@
...
+
+    //changes: changed scanf to fgets, preventing an error if input is not a number
+    fgets(binary_number, 100, stdin);
+
+    //this assigns the variable number to the converted integer
+    number = strtol(binary_number, &binary_number_ptr, 0);
```
Spikatrix: There's no need for `binary_number_ptr` as it is not used. You can pass `NULL` instead.

PR #243, opened 2019-07-10
fix the stack.c

megahertz66:
Thanks for giving me the opportunity for the first PR. I think I think I have solved the problem of on free.

### DIFF
--- a/data_structures/stack/stack.c
+++ b/data_structures/stack/stack.c
@@ -42,7 +42,7 @@ void initStack()
     grow: increases the stack by 10 elements.
           This utility function isn't part of the public interface
 */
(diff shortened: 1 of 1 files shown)


data_structures/stack/stack.c
```diff
@@ -42,7 +42,7 @@ void initStack()
     grow: increases the stack by 10 elements.
           This utility function isn't part of the public interface
 */
-void grow()
+void  grow(void ** oldMemory)
 {
```
danghai: I think you do not need `void ** oldMemory`
just free tmp

megahertz66: Yes， I will  change it.

PR #251, opened 2019-08-10
Add gnome sort

gbrunofranco:
I've cleaned up my previous PR.

### DIFF
--- a/sorting/gnome_sort.c
+++ b/sorting/gnome_sort.c
@@ -1,0 +1,46 @@
+#include <stdio.h>
+#include <stdlib.h>
+
+void sort(int *numbers, int size)
+{
+    int pos = 0;
+    while (pos < size)
+    {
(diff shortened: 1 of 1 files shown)


stepfencurryxiao:
The code reported an error on dev c++:
```
[Error]loop initial declarations are only allowed in C99 or C11 mode
[Note] use option -std=c99, -std=gnu99, -std=c11 or -std=gnu11 to compile your code
```
You can't declare variable 'i' in the 'for' loop
```c
int i;
for(i=0;i<n;i++)
{
   ......;
}
```
In this way, the 'for' loop can be compiled and passed by dev c++.

gbrunofranco:
I've fixed the problem, let me know if I have to change something else.

stepfencurryxiao:
Ok,Merge!!!

PR #252, opened 2019-08-10
Fix for loop in gnome_sort to make it C89-compatible

gbrunofranco:
Now they should all be C89-compatible.


stepfencurryxiao:
thanks @gbrunofranco

PR #255, opened 2019-08-26
fix bug

shellhub:
`binarySearch` should be returned when `left` > `right`


shellhub:
@danghai @christianbender @AnupKumarPanwar @theycallmemac

shellhub:
@stepfencurryxiao

stepfencurryxiao:
Hello,@shellhub! I also think we need to return the value when left > right. Thank you for finding this bug.

PR #256, opened 2019-09-14
Added BFS algorithm

PalAditya:
Added the **BFS** algorithm in **graphs** section, which seems very lacking. Please review it and merge if found useful 😁 

The Sample I/O from an execution is as below:

Run


danghai:
Hi @PalAditya , thank you. Merge!

PR #258, opened 2019-09-17
Added Huffman coding, Hash Sort algorithm and a maze solver

mrinal-pandey:


ashwek:
@mrinal-pandey please add comments

mrinal-pandey:
Found my implementation of Huffman to be wrong. Would update and create a new request. Thanks :)

PR #264, opened 2019-09-29
devlali numbers

isykasle:
Devlali is number that cannot be written
as the sum of any other smaller natural number i
and the individual digits of i.

### DIFF
--- a/misc/devlali
+++ b/misc/devlali
@@ -1,0 +1,39 @@
+/*
+Ioannis Sykas Leriotis  29/09/19
+
+
+Devlali is number that cannot be written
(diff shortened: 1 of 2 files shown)


ashwek:
@isykasle please add **.c** extension to your file

ashwek:
according to [Devlali number - wikipedia](https://en.wikipedia.org/wiki/Self_number) 

108, 110, 121, 132, 143, 154

these numbers **are** *Devlali number*.
But according to your program, these number **are not** *Devlali number*.

kvedala:
@isykasle are you working on fixing the errors reported above? Thanks

PR #265, opened 2019-09-29
kaprekars

isykasle:
6174 is the Kaprekar Constant. This number is special as we always get this number when following steps are followed 
for any four digit number such that all digits of number are not same.

Sort four digits in ascending order and store result in a number “asc”.
Sort four digits in descending order and store result in a number “desc”.
Subtract number larger number from smaller number, i.e., abs(asc – desc).
Repeat above three steps until the result of subtraction doesn’t become equal to the previous number.

### DIFF
--- a/misc/kaprekars
+++ b/misc/kaprekars
@@ -1,0 +1,105 @@
+/*
+Ioannis Sykas Leriotis 29/09/2019
+6174 is the Kaprekar Constant. This number is special as we always get this number when following steps are followed 
+for any four digit number such that all digits of number are not same.
+
+Sort four digits in ascending order and store result in a number “asc”.
+Sort four digits in descending order and store result in a number “desc”.
+Subtract number larger number from smaller number, i.e., abs(asc – desc).
+Repeat above three steps until the result of subtraction doesn’t become equal to the previous number.
+*/
+
+# include <stdio.h>
+# include <stdbool.h>
+# include <math.h>
+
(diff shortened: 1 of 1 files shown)


ashwek:
@isykasle please add **.c** extension to your file

misc/kaprekars
```diff
@@ -0,0 +1,105 @@
...
+Sort four digits in descending order and store result in a number “desc”.
+Subtract number larger number from smaller number, i.e., abs(asc – desc).
+Repeat above three steps until the result of subtraction doesn’t become equal to the previous number.
+*/
+
+# include <stdio.h>
+# include <stdbool.h>
+# include <math.h>
+
+int kaprekarRec(int n, int prev)
```
ashwek: what is the need to pass  **prev** as arguments if you are assigning it a different value in the function _(Line 23)_ ?

misc/kaprekars
```diff
@@ -0,0 +1,105 @@
...
+                }
+            }
+
+        }
+    int desc = 0;
+    for (i=0; i<4; i++){
+       desc = desc*10 + digits[i];
+    }
+
+    // Sort all four dgits in ascending order
```
ashwek: you have already sorted the digits in descending order in ``digits`` array. Instead of again sorting the array in ascending order, you could just iterate through the array in reverse to get the ``asc``

PR #266, opened 2019-09-30
Exponential Search added

sayanmondal2098:


searching/ExponentialSearch.c
```diff
@@ -0,0 +1,65 @@
...
+		return 0; 
+
+	// Find range for binary search by 
+	// repeated doubling 
+	int i = 1; 
+	while (i < n && arr[i] <= x) 
+		i = i*2; 
+
+	// Call binary search for the found range. 
+	return binarySearch(arr, i/2, min(i, n), x);
```
ashwek: ``min`` is not defined for c. Please define ``min``.

searching/ExponentialSearch.c
```diff
@@ -0,0 +1,65 @@
...
+} 
+
+// Driver code 
+int main(void) 
+{ 
+int arr[] = {2, 3, 4, 10, 40}; 
+int n = sizeof(arr)/ sizeof(arr[0]); 
+int x = 10; 
+int result = exponentialSearch(arr, n, x); 
+(result == -1)? printf("Element is not present in array")
```
ashwek: please print the array, & the value that was searched before printing the result. & maybe take input from users instead of using a fixed array.

PR #270, opened 2019-10-01
[added] lerp, with both an unprecise and a precise option.

vicenteferrari:
Added a simple linear interpolation algorithm and sample main function.


vicenteferrari:
bump

PR #281, opened 2019-10-01
Update binary2octal.c

AXECODER:


stepfencurryxiao:
Hello@AXECODER!  i++  is equivalent to i=i+1. I think using i + + is better than i = i + 1 because it makes the code more concise.

PR #299, opened 2019-10-02
Added conversions/hexa_to_decimal.c

dwij2812:


ashwek:
**hexa_to_binary.c** I don't think this is the ideal way to convert hexadecimal to binary.
**hexa_to_decimal.c** file name says **hexa_to_decimal** but the program is converting _hex to bin_.

PR #300, opened 2019-10-02
Add files via upload

himanshututeja1998:
Memoized FibonnaciDP.c

### DIFF
--- a/misc/FIB.c
+++ b/misc/FIB.c
@@ -1,0 +1,20 @@
+#include <stdio.h>
+#include <stdlib.h>
+void main(int argc, char *argv[])
+{
+	int n,i;
+	
+	printf("ENTER THE N");
+	scanf("%d" ,&n);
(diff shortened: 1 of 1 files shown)


misc/FIB.c
```diff
@@ -0,0 +1,20 @@
...
+#include <stdlib.h>
+void main(int argc, char *argv[])
+{
+	int n,i;
+	
+	printf("ENTER THE N");
+	scanf("%d" ,&n);
+	int fib[n];
+	fib[0]=1;
+	fib[1]=1;
```
ashwek: if users inputs __1__ _(or any number less than 1)_, then the size of __fib__ is <=1 
```
fib[0] = 0;
fib[1] = 1;
```
and these 2 lines may cause error.

himanshututeja1998: Yes sure i will change it and add it .Thanks xD

PR #306, opened 2019-10-02
added AVL tree in C

sidag99:
Added AvlTree to data structures


ashwek:
@sidag99 thank you for contributing, but we already have an implementation of AVL tree.
Closing PR.

PR #307, opened 2019-10-02
Simple bubble sort

aman566:
bubble sort Written in c


ashwek:
@aman566 thank you for contributing but a very similar version of [BubbleSort.c](https://github.com/TheAlgorithms/C/blob/master/sorting/BubbleSort.c) already exists. Closing PR.

PR #309, opened 2019-10-02
Update more leetcode solution

jonohein:
Fixed a couple which were out of order and added #1 which was missing.


danghai:
Thanks! Merge

PR #310, opened 2019-10-02
7.c :reverse of a number with overflow check added

SuranjanDaw:


leetcode/src/7.c
```diff
@@ -0,0 +1,35 @@
+int reverse(int x){
```
danghai: So many checking here. I have a good solution for it:
```
int reverse(int x){
    long res = 0;
    while (x != 0)
    {
        int tail = x % 10;
        res = res * 10 + tail;
        /* Check overflow */
        if (res > INT_MAX || res < INT_MIN)
            return 0;
        x = x / 10;
    }
    return (int)res;
}
```

leetcode/src/9.c
```diff
@@ -0,0 +1,50 @@
+
+
```
danghai: a better solution:
```
int reverse(int x){
    long res = 0;
    while (x != 0)
    {
        int tail = x % 10;
        res = res * 10 + tail;
        /* Check overflow */
        if (res > INT_MAX || res < INT_MIN)
            return 0;
        x = x / 10;
    }
    return (int)res;
}

bool isPalindrome(int x) {
    if (x < 0)
        return 0;
    else
        return x == reverse(x);
}
```

leetcode/src/7.c
SuranjanDaw: Here you are using long int. What if the system we are using, supports up to 32-bit? What if the OS does not support long int?

danghai: The question is not limit only for 32-bit.

SuranjanDaw: Yeah if we don't put that restriction then it will work. So do you want me to change and re-push?

danghai: You can put the comment on your way, it works only for 32-bit system. And then adding my solution below. Thank you

SuranjanDaw: Committed the changes. Thanks

Issue #249, opened 2019-08-08
Add more solution for Leetcode algorithm.


danghai:
I create leetcode directory. You can contribute solution for Leetcode algorithm by C language.
Good practice coding :)
Please put some comment for explanation. It helps other people understand your algorithm better

SuranjanDaw:
Added 7.c: Integer reversal with 32-bit overflow checker  and 9.c: Pallidrome without string conversion

shubhamdpatil:
Added Leetcode solutions for problem no: 2, 4, 7, 8, 21, 28
PR: #349

Kunal614:
Unique Number of Occurrences 1207
https://github.com/TheAlgorithms/C/pull/422

kvedala:
This is an excellent collection 👍 

Can you please convert the codes to independent executable programs with documentation in sync with the rest of the repo? Thanks

BhanukaC:
#703

BhanukaC:
Add LeetCode Problem 1486 using C #703

BhanukaC:
> I create leetcode directory. You can contribute solution for Leetcode algorithm by C language.
> Good practice coding :)
> Please put some comment for explanation. It helps other people understand your algorithm better

Add LeetCode Problem 1486 using C #703

BhanukaC:
Add LeetCode Problem 1486 using C #703

BhanukaC:
LeetCode 961 Problem Using C #702

BhanukaC:
Create 1295.c #701

itsvinayak:
hey @danghai, how to add an alternate algorithm for the same question, should it be added to the same file or by creating another file, what should be the syntax for naming file?

thank you,
@itsvinayak

kvedala:
> hey @danghai, how to add an alternate algorithm for the same question, should it be added to the same file or by creating another file, what should be the syntax for naming file?
> 
> thank you,
> @itsvinayak

[Contribution Guidelines](https://github.com/TheAlgorithms/C/blob/master/CONTRIBUTING.md)
Meanwhile, the Leetcode folder is currently unstructured and there is an effort to normalize the folder with respect to the repo #733

DhruvPasricha:
Hi @danghai, I would like to work on this issue,.
Can i contribute by adding some solutions ?

Panquesito7:
> Can I contribute

Thank you for your interest in contributing. 👍
Before submitting a PR, please ensure the following:

[...]

PR #311, opened 2019-10-03
249 add maximum number of ballons

rodrigoAMF:
As requested on https://github.com/TheAlgorithms/C/issues/249

Add solution to problem 1189 Maximum Number of Ballons available at: https://leetcode.com/problems/maximum-number-of-balloons/

Also updated README from leetcode folder.

### DIFF
--- a/leetcode/src/1189.c
+++ b/leetcode/src/1189.c
@@ -1,0 +1,48 @@
+int maxNumberOfBalloons(char * text){
+    /* 
+        0 -> b, 
+        1 -> a,
+        2 -> l,
+        3 -> l,
+        4 -> o,
+        5 -> o,
+        6 -> n
+    */
+    int count_letters[7] = {0};
+    
+    /* Count the number of each letter */
+    for(int i = 0; text[i] != '\0'; i++){
+        if(text[i] == 'b') {
+            count_letters[0]++;
+        }else if(text[i] == 'a') {
+            count_letters[1]++;
+        }else if(text[i] == 'l') {
+            count_letters[2]++;
+            count_letters[3]++;
+        }else if(text[i] == 'o') {
+            count_letters[4]++;
+            count_letters[5]++;
+        }else if(text[i] == 'n') {
+            count_letters[6]++;
+        }
+    }
+    
+    /* Divide by 2 the repeted letters */
+    count_letters[2] /= 2;
+    count_letters[3] /= 2;
+    count_letters[4] /= 2;
+    count_letters[5] /= 2;
+    
+    /* Max number of times which we can write ballon is equal to min value of letters on count_letter */
+    int min_counter_ballons;
+    for(int i = 0; i < 7; i++){
+        if(i == 0){
+            min_counter_ballons = count_letters[i];
+        }else if(count_letters[i] < min_counter_ballons){
+            min_counter_ballons = count_letters[i];
+        }
+    }
+    
+    return min_counter_ballons;
+}
+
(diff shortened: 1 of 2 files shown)


leetcode/src/1189.c
```diff
@@ -0,0 +1,48 @@
...
+    /* Divide by 2 the repeted letters */
+    count_letters[2] /= 2;
+    count_letters[3] /= 2;
+    count_letters[4] /= 2;
+    count_letters[5] /= 2;
+    
+    /* Max number of times which we can write ballon is equal to min value of letters on count_letter */
+    int min_counter_ballons;
+    for(int i = 0; i < 7; i++){
+        if(i == 0){
```
danghai: I think we do not need the `if-condition` here. We just initialize the variable `min_counter_ballons = count_letters[0]` outside the loop

leetcode/src/1189.c
```diff
@@ -0,0 +1,48 @@
...
+            count_letters[3]++;
+        }else if(text[i] == 'o') {
+            count_letters[4]++;
+            count_letters[5]++;
+        }else if(text[i] == 'n') {
+            count_letters[6]++;
+        }
+    }
+    
+    /* Divide by 2 the repeted letters */
```
danghai: We do not need to count 2 times for repeated character.

leetcode/src/1189.c
```diff
@@ -0,0 +1,48 @@
...
+        3 -> l,
+        4 -> o,
+        5 -> o,
+        6 -> n
+    */
+    int count_letters[7] = {0};
+    
+    /* Count the number of each letter */
+    for(int i = 0; text[i] != '\0'; i++){
+        if(text[i] == 'b') {
```
danghai: I think it is better to use *ptr for traversing the string

danghai:
Hi @rodrigoAMF . It is a good solution with explanation. I have some enhancement.
```
int maxNumberOfBalloons(char * text){
    /* 
        0 -> b, 
        1 -> a,
        2 -> l,
        3 -> o,
        4 -> n
    */
    int count_letters[5] = {0};
    int i, min_counter_ballons;
    
    for (char *ptr = text; *ptr; ptr++) {
        if(*ptr == 'b') {
            count_letters[0]++;
        }else if(*ptr == 'a') {
            count_letters[1]++;
        }else if(*ptr == 'l') {
            count_letters[2]++;
        }else if(*ptr == 'o') {
            count_letters[3]++;
        }else if(*ptr == 'n') {
            count_letters[4]++;
        }
    }
    
    /* Divide by 2 the repeted letters */
    count_letters[2] /= 2;
    count_letters[3] /= 2;
    
    /* Max number of times which we can write ballon is equal to min value of letters on count_letter */
    min_counter_ballons = count_letters[0];
    for(i = 1; i < 5; i++){
        if(count_letters[i] < min_counter_ballons)
            min_counter_ballons = count_letters[i];
    }
    
    return min_counter_ballons;
}
```

leetcode/src/1189.c
rodrigoAMF: Yes good point thanks!

leetcode/src/1189.c
rodrigoAMF: With that modification the code seems more readable but just for curiosity does this make any difference in terms of optimization?

leetcode/src/1189.c
rodrigoAMF: That's true thanks for pointing out!

rodrigoAMF:
Hi @danghai thanks for your time reviewing my code! I've updated the code with your enhancements.

leetcode/src/1189.c
danghai: Ah it does not have any optimization. It is just more readable, and safe.

danghai:
Thank you. I hope you can contribute more leetcode problem.

PR #314, opened 2019-10-03
Added multikey quicksort

JianLoong:
Wikipedia entry can be found [here](https://en.wikipedia.org/wiki/Multi-key_quicksort).

Official paper can be found [here.](https://www.cs.princeton.edu/~rs/strings/paper.pdf)

Based on demo program [here.](https://www.cs.princeton.edu/~rs/strings/demo.c)

### DIFF
--- a/sorting/multikey_quicksort.c
+++ b/sorting/multikey_quicksort.c
@@ -1,0 +1,407 @@
+/* demo.c -- Implementations of multikey quicksort and ternary search trees
+   Usage
+    demo                  Run basic timings on /usr/dict/words
+    demo <file>           Run basic timings on <file>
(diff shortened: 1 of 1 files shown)


danghai:
Wow! thank you. Merge

kvedala:
@JianLoong and @danghai While maintaining the repo, I was reviewing the code and couldn't understand some parts of the code. Help would be greatly appreciated:
1. [this line](https://github.com/TheAlgorithms/C/blob/81f4428569254f08b6c8959648e0837df3850350/sorting/multikey_quick_sort.c#L259) does not make sense because `pp` is undefined at the first iteration
2. The help info at the top of the file, though nice, is not what the program does. I do no see the command-line arguments being used in the `main()` function.

PR #324, opened 2019-10-03
added insert right insert left delete right and delete left functions

harshvora7:
added insert right, insert left, delete right, delete left functions. list has dynamic memory allocation. queues, double ended queues, and stacks can be implemented in the same program.

### DIFF
--- a/data_structures/linked_list/singly_link_list_deletion.c
+++ b/data_structures/linked_list/singly_link_list_deletion.c
@@ -2,97 +2,136 @@
   It is assumed that the data in nodes will be an integer, though
   function can be modified according to the data type, easily.
   deleteNode deletes a node when passed with a key of the node.
+  The linked list can be used as a stack by implementing the functions insertLeft and deleteLeft
+  The linked list can be used as a queue by implementing the functions insertRight and deleteLeft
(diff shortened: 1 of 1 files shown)


data_structures/linked_list/singly_link_list_deletion.c
```diff
@@ -2,97 +2,136 @@
...
+    start=start->link;
+    printf("number deleted is: %d",t->info);
+    free(t);
+    return start;
+}
+///////////////////////////////////////////////////////
+NODE deleteRight(NODE start)
+{
+  NODE t=start,prev=NULL;
+  while(t->link!=NULL)
```
ashwek: is ``start`` is ``NULL``, ``t->link`` will cause error.

data_structures/linked_list/singly_link_list_deletion.c
```diff
@@ -2,97 +2,136 @@
...
+  {
+    prev=t;
+    t=t->link;
+  }
+  printf("number deleted is: %d",t->info);
+  free(t);
+  prev->link=NULL;
```
ashwek: if there is only 1 node in the linked list, ``prev`` will be ``NULL`` and ``prev->link`` will cause error

PR #325, opened 2019-10-03
Correct and improve prime test

0Zeta:
There are no prime numbers smaller than 2, but 2 itself is prime.

### DIFF
--- a/misc/Prime.c
+++ b/misc/Prime.c
@@ -2,7 +2,13 @@
 #include <math.h>
 
 int isPrime(int x) {
-    for (int i = 2; i < sqrt(x); i++) {
+    if (x == 2) {
+        return 1;
+    }
+    if (x < 2 || x % 2 == 0) {
+        return 0;
+    }
+    for (int i = 3; i < sqrt(x); i += 2) {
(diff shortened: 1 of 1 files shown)


0Zeta:
> Looks good. Although, it would be better if you move the `sqrt` to another variable so that it won't get evaluated in every iteration of the loop. Also, some spaces around the modulus operator would be nice: `if (x % i == 0)`

Implemented your suggestions

misc/Prime.c
```diff
@@ -2,8 +2,17 @@
...
+    if (x == 2) {
+        return 1;
+    }
+    if (x < 2 || x % 2 == 0) {
+        return 0;
+    }
+    
+    double squareRoot = sqrt(x);
+    
+    for (int i = 3; i < squareRoot; i += 2) {
```
ashwek: the condition should be ``i <= squareRoot``.

because in the existing code,  If ``x`` is **9**, then ``squareRoot`` becomes **3**, and the condition in the for loop fails & ``return 1;`` is executed implying that **9** is a prime number

PR #326, opened 2019-10-03
Solution to Leetcode Problem 173

SaurusXI:
In reference to issue https://github.com/TheAlgorithms/C/issues/249

### DIFF
--- a/leetcode/README.md
+++ b/leetcode/README.md
@@ -30,6 +30,7 @@ LeetCode
 |153|[Find Minimum in Rotated Sorted Array](https://leetcode.com/problems/find-minimum-in-rotated-sorted-array/) | [C](./src/153.c)|Medium|
 |160|[Intersection of Two Linked Lists](https://leetcode.com/problems/intersection-of-two-linked-lists/) | [C](./src/160.c)|Easy|
 |169|[Majority Element](https://leetcode.com/problems/majority-element/) | [C](./src/169.c)|Easy|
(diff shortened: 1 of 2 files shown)


danghai:
Nice! Thank you

leetcode/README.md
```diff
@@ -30,6 +30,7 @@ LeetCode
 |153|[Find Minimum in Rotated Sorted Array](https://leetcode.com/problems/find-minimum-in-rotated-sorted-array/) | [C](./src/153.c)|Medium|
```
Anirudha10, line 30: A simple solution is to traverse the complete array and find a minimum. This solution requires O(n) time. 
We can do it in O(Logn) using Binary Search. If we take a closer look at the above examples, we can easily figure out the following pattern:

The minimum element is the only element whose previous is greater than it. If there is no previous element, then there is no rotation (the first element is minimum). We check this condition for the middle element by comparing it with (mid-1)’th and (mid+1)’th elements.
If the minimum element is not at the middle (neither mid nor mid + 1), then the minimum element lies in either the left half or right half. 
If the middle element is smaller than the last element, then the minimum element lies in the left half
Else minimum element lies in the right half.
[Task 1.docx](https://github.com/TheAlgorithms/C/files/7281458/Task.1.docx)

Anirudha10: @danghai added solution to problem 153 #326

PR #327, opened 2019-10-03
Solution to leetcode Problem 190 and 191

SaurusXI:
In reference to issue https://github.com/TheAlgorithms/C/issues/249


leetcode/src/190.c
```diff
@@ -0,0 +1,10 @@
+uint32_t reverseBits(uint32_t n) {
+    uint TotalBits = 32;
+    uint32_t reverse_int = 0;
+    uint i;
+    for(i = 0; i < TotalBits; i++) {
+        if((n & (UINT32_C(1) << i)))
```
danghai: Could you put some comment for explanation? Ex: `check bit is 1 or 0`

SaurusXI: Done. Please take a look

PR #330, opened 2019-10-04
added hexal_to_octal in conversions

anon6405:


conversions/hexal_to_octal.c
```diff
@@ -0,0 +1,134 @@
+**
+ * C program to convert Hexadecimal to Octal number system
+ */
```
ashwek: there is no multi-line comment starting symbol ``/*``

ashwek:
@anon6405 i don't think this is an ideal way to convert hexadecimal to octal

PR #335, opened 2019-10-04
Added ArmstrongNumber.c

Nikhil-Bhat6:
Added a solution for the popular question - ArmstrongNumber

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,19 +73,20 @@ C
 
 
 ## Misc
+	- ArmstrongNumber	
 	- Binning
 	- Factorial
 	- Fibonacci
+	- Greatest Common Divisor	
 	- isArmstrong
 	- LongestSubSequence
 	- palindrome
+	- prime factorization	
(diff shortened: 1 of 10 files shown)


misc/ArmstrongNumber.c
```diff
@@ -0,0 +1,64 @@
+//A number is called as Armstrong number if sum of cubes of digits of number is equal to the number itself.
```
ashwek: the definition is not perfect. Please change it to something like

> Armstrong number is a number that is the sum of its own digits each raised to the power of the number of digits. 

[Narcissistic Number - wikipedia](https://en.wikipedia.org/wiki/Narcissistic_number)

README.md
```diff
@@ -73,19 +73,20 @@ C
 
 
 ## Misc
+	- ArmstrongNumber
```
ashwek: ``isArmstrong`` exists on line 81

PR #342, opened 2019-10-05
Added Conversition and evaluation of infix to postfix expression

ankit-tailor:
let me know if you can not understand anything

### DIFF
--- a/data_structures/stack/evaluation_of_postfix.c
+++ b/data_structures/stack/evaluation_of_postfix.c
@@ -1,0 +1,134 @@
+#include<stdio.h>
+#include<string.h>
+#include<stdlib.h>
+#include<math.h>
+
+struct Stack
+{
+    int top;
+    unsigned capacity;
+    int* array;
+    /* data */
+};
+struct Stack* createStack(unsigned capacity)
+{
+    struct Stack* stack = (struct Stack*) malloc(sizeof(struct Stack));
+    if(!stack)
+     return NULL;
+    stack->top = -1;
+    stack->capacity = capacity;
+    stack->array = (int*) malloc(sizeof(int));
+    if(!stack->array)
+        return NULL;
+    return stack;
+}
+char peek(struct Stack* stack)
+{
+    return stack->array[stack->top];
+}
+void push(struct Stack* stack, char val)
+{
+    stack->array[++stack->top] = val;
+}
+int isEmpty(struct Stack* stack)
+{
+    if(stack->top==-1)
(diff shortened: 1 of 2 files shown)


data_structures/stack/infixToPostfix.c
```diff
@@ -0,0 +1,100 @@
...
+struct Stack* createStack(unsigned capacity)
+{
+    struct Stack* stack = (struct Stack*) malloc(sizeof(struct Stack));
+    if(!stack)
+        return NULL;
+    stack->top = -1;
+    stack->capacity = capacity;
+    stack->array = (int*) malloc(stack->capacity * sizeof(int));
+    if(!stack->array)
+        return NULL;
```
ashwek: deallocate memory allocated to ``stack`` before returning to avoid memory leaks.

data_structures/stack/infixToPostfix.c
```diff
@@ -0,0 +1,100 @@
+#include<stdio.h>
+#include<string.h>
+
+struct Stack
+{
+    int top;
+    unsigned capacity;
+    int* array;
```
ashwek: I think ``array`` should be of type ``char`` because it is used to store operators (which are characters) to avoid unnecessary implicit type conversion in ``pop`` and ``push``.

data_structures/stack/infixToPostfix.c
```diff
@@ -0,0 +1,100 @@
...
+            return 1;
+        case '*':
+        case '/':
+            return 2;
+        case '^':
+            return 3;
+    }
+    return -1;
+}
+int infixToPostfix(char* exp)
```
ashwek: change the return type to ``void`` if nothing is returned

data_structures/stack/evaluation_of_postfix.c
```diff
@@ -0,0 +1,134 @@
...
+        {
+            val1 = pop(stack);
+            val2 = pop(stack);
+            switch (exp[i])
+            {
+                case '+': push(stack, val1+val2); break;
+                case '-': push(stack, val1-val2); break;
+                case '*': push(stack, val1*val2); break;
+                case '/': push(stack, val1/val2); break;
+                case '^': push(stack, pow(val2, val1)); break;
```
ashwek: If input is _9-5_, correct answer is _4_. But yout code outputs _-4_
the equation should be **operand2**  **_operator_**   **operand1**
``val2 + val1``. not ``val1 + val2``

PR #345, opened 2019-10-05
Add Stooge Sort

ZykielErx:
Revised Readme.md to include stooge sort on the list
Added sorting/StoogeSort.c


ashwek:
@ZykielErx thank you for contributing, but we already have an implementation of stooge sort. Closing PR.

PR #346, opened 2019-10-05
Add Tim Sort

ZykielErx:
Revised Readme.md to include Tim sort on the list
Added sorting/TimSort.c

### DIFF
--- a/sorting/TimSort.c
+++ b/sorting/TimSort.c
@@ -1,0 +1,127 @@
+// Sorting of array list using Tim Sort 
+#include <stdio.h>
+#include <stdlib.h>
+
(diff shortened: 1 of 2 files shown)


sorting/TimSort.c
```diff
@@ -0,0 +1,127 @@
...
+    for (int i = 0; i < n; i++)
+        printf("%d ", arr[i]);
+    printf("\n");
+}
+
+// Driver program to test above function
+int main()
+{
+    int n;  // Size of array elements
+    int arr[n];
```
ashwek: since ``n`` is not initialized, it would have junk value. And ``arr`` will be of random size.
Declare ``arr`` after taking input into ``n``.

PR #348, opened 2019-10-05
Create Infix_to_Postfix.C

Chayan-19:
C program to convert infix expression to postfix expression using stack.


data_structures/stack/Infix_to_Postfix.C
```diff
@@ -0,0 +1,133 @@
...
+	if (!stack) 
+		return NULL; 
+
+	stack->top = -1; 
+	stack->capacity = capacity; 
+
+	stack->array = (int*) malloc(stack->capacity * sizeof(int)); 
+
+	if (!stack->array) 
+		return NULL;
```
ashwek: deallocate memory allocated to ``stack`` before returning to avoid memory leaks.

data_structures/stack/Infix_to_Postfix.C
```diff
@@ -0,0 +1,133 @@
...
+#include <stdio.h> 
+#include <string.h> 
+#include <stdlib.h> 
+
+// Stack type 
+struct Stack 
+{ 
+	int top; 
+	unsigned capacity; 
+	int* array;
```
ashwek: I think ``array`` should be of type ``char`` because it is used to store operators (which are characters) to avoid unnecessary implicit type conversion in pop and push.

PR #350, opened 2019-10-05
Add PID (Proportional Integral Derivative) Controller

ciaranlangton:
This PR adds the PID Controller - a common (and likely most ubiquitous) linear control systems algorithm. For now, since there are no other control algorithms present in the repo, I have added it into the `misc` folder. Apologies if this is slightly out of scope of your definition of algorithms, but I figured if the client server socket example was in scope, then this likely would be too.

More information about the PID Controller can be found here: https://en.wikipedia.org/wiki/PID_controller

Thanks, Ciaran

### DIFF
--- a/misc/pid.c
+++ b/misc/pid.c
@@ -1,0 +1,80 @@
+/**
+ * PID Controller
+ * 
+ * The PID controller is a linear control algorithm that has three terms:
+ *  - Proportional: A simple scaling of the error value by a gain kP
+ *  - Integral: Integration of the error value over time, then multipled by gain kI
+ *  - Derivative: Rate of change of the error value over time, multiplied by gain kD
+ * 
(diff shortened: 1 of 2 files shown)


misc/pid.c
```diff
@@ -0,0 +1,78 @@
...
+    printf("Using kP: %f, kI: %f, kD: %f\n", controller.kP, controller.kI, controller.kD);
+
+    // How often the pid_step algorithm expects to be called. In a real life scenario this would
+    // be provided by calling time(NULL) - last_time, or by calling the function reliably at X Hz (using a timer or RTOS etc)
+    // For demonstration of this algorithm though, it is defined below as 1 second, allowing easy testing of integral
+    // and derivative terms.
+    float time_step = 1;
+
+    float error_value;
+    while (1) {
```
ashwek: please provide a way for user to exit the loop.

PR #351, opened 2019-10-05
Contribution to the Algorithm Master

Anushka-codergirl:
Added a few programs related to tree algorithm in C language.


tree/DFS.c
```diff
@@ -0,0 +1,116 @@
...
+            prev->left = temp;
+        }
+    }
+}
+ 
+void DFS(struct node *head)
+{
+    if (head)
+    {
+        if (head->left)
```
ashwek: we first print the content of the node, & then traverse to the lower levels

ashwek:
@Anushka-codergirl what is the purpose / functioning of **Set1.c** ??

ashwek:
@Anushka-codergirl thank you for contributing, but we already have an implementation of Binary Search Tree. Please remove the **BINSearchTree.c** file from the PR & instead of adding a new file for the same program it would be great if you could add features to the existing file.

PR #355, opened 2019-10-06
Added Cartesian to Polar coordinates algorithm

Cornul11:


misc/cartesianToPolar.c
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+give as arguments to the executable two x and y coordinates
+outputs a polar coordinate
+**/
+int main() {
+    double x, y;
+    double r, theta, thetaFinal;
+    scanf("%lf %lf", &x, &y);
+    r = hypot(x, y);
```
kvedala, line 14: This code, from line# 15 to line 42 - 90% of the code is useless. All the if-else cases are simply void. 
only line# 14 and 43 are needed.

PR #367, opened 2019-10-07
Added : conversion of Decimal Integer to BCD

KaProDes:
Any Integer can be converted to it's BCD equivalent.
Ex : If 22 is entered, the BCD quivalent is 0010_1001 i.e. 39 in decimal.
- The program will therefore output:
- TrueBCD form : 101001
- BCD form's decimal equivalent : 39


ashwek:
@SnowyKay thank you for contributing, but there seems to be a problem, previously accepted commits appear in your commits section too. So i would suggest you to please close this PR & create a new PR.

PR #369, opened 2019-10-07
Add Code of Deletion and Sorting of Nodes in Binary Tree

GarvTambi:


ashwek:
@GarvTambi thank you for contributing, but we already have an implementation of binary search tree (with node deletion). So please remove _Deletion of node in Binary Tree_ from the PR. It would be great if you could add some functionality to the already existing program.
Please add **.c** extension to the _Binary tree sort_ file.

There seems to be some problem, previously accepted commits appear in your commits section too. So i would suggest you to close this PR, & please create a new PR with the **Binary tree sort.c** file.

PR #370, opened 2019-10-07
Add more solution for leetcode

jonohein:
Added #5 for leetcode, fixed typo in leetcode/readme

### DIFF
--- a/leetcode/src/5.c
+++ b/leetcode/src/5.c
@@ -1,3 +1,11 @@
+int checkPalindrome(char *s, int start, int end) {
+    for(int k = 1; k <= (end-start)/2; k++) {
+        if(*(s + start + k) != *(s + end - k))
+            return 0;
+    }
+    return 1;
+}
+
 char * longestPalindrome(char * s){
 	char *result = NULL;
 	int strSize = strlen(s);
@@ -7,13 +15,7 @@ char * longestPalindrome(char * s){
 	for(int i = 0; i < strSize; i++){
 		for(int j = i + largest; j < strSize; j++){
 			if(*(s+i) == *(s+j)){
-				int pass = 1;
-				for(int k = 1; k <= (j-i)/2; k++){
-					if(*(s+i+k) != *(s+j-k)){
-						pass = 0;						
-						break;
-					}
-				}
+				int pass = checkPalindrome(s, i, j);
 				if(pass && largest < j-i+1){
 					largest = j - i + 1;
 					start = i;
--- a/leetcode/src/5.c
+++ b/leetcode/src/5.c
@@ -1,3 +1,4 @@
+/* Approach #1 */
 int checkPalindrome(char *s, int start, int end) {
     for(int k = 1; k <= (end-start)/2; k++) {
         if(*(s + start + k) != *(s + end - k))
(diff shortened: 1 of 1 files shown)


leetcode/src/5.c
```diff
@@ -0,0 +1,34 @@
+char * longestPalindrome(char * s){
+	char *result = NULL;
+	int strSize = 0;
```
danghai: You can use func `strlen` instead of for-loop:
`int strSize = strlen(s);`

leetcode/src/5.c
```diff
@@ -0,0 +1,34 @@
...
+		strSize = i + 1;
+	}
+	int start = 0;
+	int end = 0;
+	int largest = 0;
+	for(int i = 0; i < strSize; i++){
+		for(int j = i + largest; j < strSize; j++){
+			if(*(s+i) == *(s+j)){
+				int pass = 1;
+				for(int k = 1; k <= (j-i)/2; k++){
```
danghai: It is nice when you can split it by another function (ex: `bool checkPalindrome(char *s, int start, int end)`)?

danghai: I mean:
```c

int checkPalindrome(char *s, int start, int end) {
    for(int k = 1; k <= (end-start)/2; k++) {
        if(*(s + start + k) != *(s + end - k))
            return 0;
    }
    return 1;
}
char * longestPalindrome(char * s){
	char *result = NULL;
	int strSize = strlen(s);
	int start = 0;
	int end = 0;
	int largest = 0;
	for(int i = 0; i < strSize; i++){
		for(int j = i + largest; j < strSize; j++){
			if(*(s+i) == *(s+j)){
				int pass = checkPalindrome(s, i, j);
				if(pass && largest < j-i+1){
					largest = j - i + 1;
					start = i;
					end = j;
				}
			}
		}
	}
	result = (s+start);
	// Destructive
	if(largest > 0 && end+1 != strSize+1){
		s[end+1] = '\0';
	}
	return result;
}
```

danghai:
Hi @jonohein . I gave my feedback. Could you add my way bellow your code? Thank you
```c

int maxVal(int a, int b) {
    return a > b ? a : b;
}

int extend(char *s, int start, int end) {
    int len = strlen(s);
    for(; start >= 0 && end < len; start--, end++) {
        if(*(s + start) != *(s + end))
            break;
    }
    return end - start - 1;
}

char * longestPalindrome(char * s){
    int max = 0, idx = 0, len = strlen(s);
    for(int i = 0; i < len; i++) {
        int len1 = extend(s, i, i);       /* For even string */
        int len2 = extend(s, i, i + 1);   /* For odd string  */
        if (max < maxVal(len1, len2)) {
            idx = (len1 > len2) ? (i - len1 / 2) : (i - len2 / 2 + 1);
            max = maxVal(len1, len2);
        }
    }
    char *res = malloc((max+1) *sizeof(char));
    memcpy(res, &s[idx], max);
    res[max] = '\0';
    return res;
}
```

leetcode/src/5.c
jonohein: Thanks, I wasn't sure if they were meant to be standalone functions or not

leetcode/src/5.c
jonohein: Yeah, that's probably better. Slipped my mind.

jonohein:
How does that look @danghai?

danghai:
It looks good! Merge

PR #373, opened 2019-10-07
Create length_linkedList.c

Souvikns:
A program that shows how to calculate the length of the linkedlist


ashwek:
@Souvikns thankyou for contributing.

In  *search_Array.c* you have implemented Linear Search, & we already have a copy of Linear Search in the _searching_ directory.

With *length_linkedList.c* instead of adding a new program only to find the length of the list, it would be great if you could add the _length_ feature in the already existing programs in the *data_structures/linked_list* directory.

PR #374, opened 2019-10-07
solved searching by recursively

aman-ku:
I wrote Binary and Linear Search by using recursive technique in C language


ashwek:
@aman-ku thank you for contributing, but we already have an recursive implementation of binary search, so please remove binary search from the PR.

There seems to be some problem, previously accepted commits appear in your commits section too. So i would suggest you to close this PR, & please create a new PR with the **LinearSearch_recursive.c** & move the file to the **sorting** directory.

PR #376, opened 2019-10-07
Added Algorithm of conversion from decimal to any base

XenomShox:


ashwek:
@XenomShox  thank you for contributing, but there seems to be some problem, previously accepted commits appear in your commits section too. So i would suggest you to close this PR, & please create a new PR with the your program.

PR #382, opened 2019-10-08
Create factsimple

codetheorem:


ashwek:
@codetheorem thank you for contributing, but we already have an implementation of factorial.
Closing PR.

PR #395, opened 2019-10-09
Added rot13.c

LucaPalumbo:
Added rot13 algorithm to cipher directory.


cipher/rot13.c
```diff
@@ -0,0 +1,49 @@
...
+void rot13(char* text);
+
+int main(int argc, char *argv[]){
+	int opt;
+	if (argc != 2){
+		usage( argv[0] );
+		return -1;
+	}
+	rot13(argv[1]);
+	return 1;
```
ashwek: we return **0** to indicate _Successful_ termination, & other values to indicate some error / failure.

PR #398, opened 2019-10-10
Create Large_Factorials.c

HrishiNarayanan:
Algorithm for calculation of very large factorials (till 5000! in this case). The code overcomes the problem that we do not have a large enough data type to store such large numbers.


ashwek:
@HrishiNarayanan please add ``printf``'s to specify what input is requested. And please add comments too.

PR #399, opened 2019-10-11
create string sorting

akshay-gupta123:
this pr is for adding string sorting in sorting folder

### DIFF
--- a/data_structures/stack/stackandqueue.c
+++ b/data_structures/stack/stackandqueue.c
@@ -1,0 +1,123 @@
+#include<stdio.h>
+#include<stdlib.h>
+ 
+
+typedef struct Node node;
+struct Node{
+    int value;
+    struct Node * next;
+};
+
+
+
+struct Node * Q1_Head;
+struct Node * Q1_tail;
+
+struct Node * Q2_Head;
+struct Node * Q2_tail;
+
+// node*top;
+
+
+node* new_node(int x) {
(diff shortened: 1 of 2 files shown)


hemanth-kotagiri (changes requested):
Hello, @akshay-gupta123. There's already a stack implementation in the same directory. I would suggest you take a look at it. I feel it's better if you can make this PR and ``implementation of queue`` specific. 

Also, about your second file, which you call ``string sorting``, please provide a description of which sorting algorithm you have used and document the code that you have written as well.

My suggestion would be to raise two individual PRs for the two files that you have combined here.

With that taken care of, always make sure to read the [contribution guidelines](https://github.com/TheAlgorithms/C/blob/master/CONTRIBUTING.md) and please abide by it.

Here are a few things I would suggest:
1. Document your code well.
2. Provide more description in your PR.
3. Follow the `typical-structure of a program` provided in the contribution guidelines.
4. follow ``snake_casing`` for your filename.


Thanks!

akshay-gupta123:
Hi, @hemanth-kotagiri first file is the implementation of stack using queue.
I will make two individual PR !!

hemanth-kotagiri:
@akshay-gupta123 now do you see the consequences of a least descriptive PR? 
That is the reason, we always want contributor to have a clear, concise description of what they've implemented.

PR #405, opened 2019-10-13
Palindrome (strings)

KN027:


misc/p.c
```diff
@@ -0,0 +1,25 @@
+#include<stdio.h>
+#include<conio.h>
```
ashwek: please avoid using ``conio.h`` _(getch)_, as it is platform specific, & wouldn't compile on linux _(gcc)_ & many other compilers.

ashwek:
@KN027 please resolve merge conflicts & please rename your program to something descriptive.

PR #414, opened 2019-10-15
adding convolution

singhshrey:
Convolution in DSP


kvedala:
could you elaborate on how this is different from #414?

PR #417, opened 2019-10-15
stack using array

vivek0206:


ashwek:
@vivek0206 thank you for contributing but we already have an implementation of stack.
Closing PR.

PR #420, opened 2019-10-16
Added OctalToDecimal Converter

VikrantReddy:


ashwek:
@VikrantReddy thank you for contributing, but please format your code properly, use either spaces or tabs, don't mix both.
You have deleted an existing file.

PR #437, opened 2019-10-20
added infix to postfix conversion for the String.

Utsav-Virani:


Utsav-Virani:
i would like to work with you.

PR #444, opened 2019-10-20
Richardson_ALGO.c

sibi812:
This is a code, which uses Richardson Algorithm to solve Matrices of the form Ax=b. If you have any doubts in the code, I would be happy to explain the code.


danghai:
Could you put some explaination for this algorithm? Thank you

PR #448, opened 2019-10-21
Project Euler Solution

sselinkurt:
Project Euler Solution for 12


kvedala:
Thank you for your contribution. But the code has been outdated with [two implementations](https://github.com/TheAlgorithms/C/blob/master/project_euler/problem_12).

PR #454, opened 2019-10-22
Add queue data structure

r00tm4k3r:


ashwek:
@r00tm4k3r thank you for contributing, but we already have an implementation of queue. It would be great if you could add functionality / improve the existing program.
Closing PR.

PR #460, opened 2019-10-23
Update LinearSearch.c

Sai-nook73:


sorting/SelectionSort.c
```diff
@@ -1,64 +1,44 @@
...
+    int temp;
     for(int i=0; i<size; i++) {
-        int min_index = i;
         for(int j=i+1; j<size; j++) {
-            if(arr[min_index] > arr[j]) {
-                min_index = j;
+            if(arr[i] > arr[j]) {
+                temp=arr[i];
+                arr[i]=arr[j];
+                arr[j]=temp;
```
ashwek: in selection sort, we choose the smallest element (for ascending order) between _i+1 to n_ & replace it with the i<sup>th</sup> element.
by swapping elements in every iteration, you have basically converted section sort to bubble sort.

PR #471, opened 2019-10-26
Nearly Sorted Algorithm.c

rageremix:
Given an array of n elements, where each element is at most k away from its target position. The task is to print array in sorted form.

### DIFF
--- a/sorting/Nearly Sorted Algorithm.c
+++ b/sorting/Nearly Sorted Algorithm.c
@@ -1,0 +1,31 @@
+#include<stdio.h>
+void sort(int a[],int n,int k)
+{
+    int key,j;
+    for(int i=1;i<n;i++)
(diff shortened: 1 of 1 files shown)


sorting/Nearly Sorted Algorithm.c
```diff
@@ -0,0 +1,31 @@
...
+        key=a[i];
+        j=i-1;
+        while(j>=0 && a[j]>key)
+        {
+            a[j+1]=a[j];
+            j=j-1;
+        }
+        a[j+1]=key;
+    }
+}
```
ashwek: what is the purpose of taking ``k`` as input if it isn't used anywhere?
``sort`` is performing normal insertion sort, & we already have an implementation of insertion sort. Closing PR.

PR #473, opened 2019-10-26
Adding leetcode Rotate Array (189.c) and Count Primes (204.c)

andre143:
Leetcode Solution:
Problem 189: Rotate Array
Problem 204: Count Primes
#249

### DIFF
--- a/leetcode/src/204.c
+++ b/leetcode/src/204.c
@@ -1,0 +1,16 @@
+int countPrimes(int n){
+    int count = 0;
+    int isPrime;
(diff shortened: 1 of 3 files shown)


leetcode/src/204.c
```diff
@@ -0,0 +1,16 @@
+int countPrimes(int n){
+    int count = 0;
+    int isPrime;
+    for (int i = 2; i <= n; i++){
```
danghai: It is fail the testcase n = 2. Expected output : 0, your output : 1

PR #481, opened 2019-10-28
Implementation of dictionary using binary trees

RaghavRoy145:
Hacktober Fest FTW


ashwek:
@RaghavRoy145 

For input,
word: a
meaning: a

word: ab
meaning: ab

there should be 2 nodes in the dictionary (tree), but your program output **Word exists!!** even though both are different words

RaghavRoy145:
Ah yes, there's probably an issue with the check function, I'll fix that
and make another request.
Thank you so much for bringing it to my notice 😊

PR #496, opened 2019-11-05
Create treetravel_norecursion.c

worangnidongdong:


worangnidongdong:
new file

ashwek:
@worangnidongdong thank you for contributing, but we already have an implementation of [Binary Tree](https://github.com/TheAlgorithms/C/blob/master/data_structures/binary_trees/binary_search_tree.c). It would be great if you could please add new functionalities / improve the existing program.
Closing PR.

PR #497, opened 2019-11-07
Indented "counting Sort" and fixed compilation errors

ubc1729:
Added description about advantage of counting sort over comparison sort and
how counting sort can be stabilized.

### DIFF
--- a/sorting/countingSort.c
+++ b/sorting/countingSort.c
@@ -1,36 +1,46 @@
-#include <stdio.h>
-#include<string.h>
 /*
   > Counting sort is a sorting technique based on keys between a specific range.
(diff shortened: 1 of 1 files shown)


ashwek:
@ubc1729 thank you for contributing. Please format code properly, use either spaces or tabs, please don't mix both.

ubc1729:
@ashwek Sure I'll edit that
But I have indented it properly on my local editor
after pushing indentation changed slightly

ashwek:
@ubc1729 because different editors handle _tab_ differently. Github substitutes every _tab_ with _2 spaces_ (i guess), maybe your editor uses some different config so the indentation looks different.

PR #499, opened 2019-11-09
HEXtoBIN

dvk65:


leetcode/src/21.c
```diff
@@ -0,0 +1,56 @@
+/* Iterative approach */
+struct ListNode* mergeTwoLists(struct ListNode* l1, struct ListNode* l2) {
+    struct ListNode *list = NULL;
+    struct ListNode *tmp = NULL;
+    
+    if (!l1)
```
kvedala: the conditions should be changed to:
```c
if(!l1 && !l2)
    return NULL;
else if (!l1)
    return l2;
else if (!l2)
    return l1;

/* here does code for the case 
if (l1 && l2)
but no need for if-else block. This eliminates a bug and also makes it more readable */
```

PR #500, opened 2019-11-28
GitHub Action run cpplint only on files that differ from master

cclauss:
GitHub Action that runs cpplint only on those files in a pull request that have been modified vs. origin/master
This allows for gradual compliance with cpplint as only files added or modified are checked.
Other optional filepath verifications are commented out at the end of this file.


kvedala:
Thank you @cclauss. The existing code does not conform to the CPPLINT standard and would break the system. I suggest we could use the `clang-format` program to automate the formatting once on all the existing code and the new ones instead of just throwing in an error.

kvedala:
@cclauss wouldn't this PR be superseded by #543 ?

PR #507, opened 2020-01-03
heapsort does not work for sorted input 1,2,3,4,5

ericcurtin:
So I copied the C++ version from TheAlgorithms and re-wrote it in C.
Now it works.


danghai:
@ericcurtin . Thanks. Merge

PR #512, opened 2020-01-28
Negative index of array

singla007:
It prevents array[-1] index as negative index doesn't exists in array


kvedala:
@singla007 can you please resolve the merge conflicts and test the code. thanks.

PR #513, opened 2020-02-02
index now starts from 1

tania-cmyk:
Please review and suggest changes if my solution doesn't take care of the issue.

### DIFF
--- a/sorting/gnome_sort.c
+++ b/sorting/gnome_sort.c
@@ -6,17 +6,18 @@ void sort(int *numbers, int size)
     int pos = 0;
     while (pos < size)
     {
-        if (numbers[pos] >= numbers[pos-1])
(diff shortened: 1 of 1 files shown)


sorting/gnome_sort.c
```diff
@@ -6,17 +6,18 @@ void sort(int *numbers, int size)
     int pos = 0;
     while (pos < size)
     {
-        if (numbers[pos] >= numbers[pos-1])
+       if (pos == 0)
+          pos = 1;
+        if (numbers[pos] >= numbers[pos-1] && pos==0)
```
kvedala, line 11: will the case `pos==0` be ever true after the change?

kvedala:
Since the question has not been answered, a revert pull-request has been created

Issue #487, opened 2019-10-28
Shortest path 


Shreya2704:
To find the shortest path between 2 cities

sagnik-chatterjee:
Hi can i take up this issue ?
Plus any specific algortihm you have in mind ?

sagnik-chatterjee:
hi @Shreya2704  can you verify if the issue is resolved ? I have done a pull request for this ...if you find this solves your problem kindly merge it or if further additions need to be made ,please list them here.

deadshotsb:
@sagnik-chatterjee Well the Djikstra's algorithm can be used to find a single source shortest path, but as "To find the shortest path between 2 cities", I find it more like applying an all pair Shortest path (floyd warshall algorithm) You may consider adding a dynamic programming approach for that, you may refer to: https://www.geeksforgeeks.org/floyd-warshall-algorithm-dp-16/

ManavpreetSingh:
Can i take this issue ? Let me know if you have any particular choice !

plongueira:
Maybe use discrete mathematics theory for example directed graphs and weighted graphs have could solve the problem (i.e) the distance between 2 cities without a direct connection is unknown so in order to get a close reflect of the reality you could check different paths between those two cities,then adding the weights between each pair of cities from city source to destiny and that'd be all

deadshotsb:
@ManavpreetSingh Yes sure

HybridDog:
In practice, e.g. in a car navigation system, the graph is preprocessed so that the distance between two cities can later be queried in e.g. O(sqrt(n) log(n)), which is a lot faster than Dijkstra's algorithm (excluding the preprocessing time).
https://en.wikipedia.org/wiki/Contraction_hierarchies

PR #514, opened 2020-02-02
added file for djikstars's in /greedy_approach

sagnik-chatterjee:
added djikstra's algorithm in c , solves issue #487


deadshotsb:
Good work but once refer to #487

PR #515, opened 2020-02-12
Improve space complexity from O(n) to O(1)

rcmarc:


misc/Fibonacci_DP.c
```diff
@@ -25,6 +25,8 @@ int fib(int n)
   for (i = 2; i <= n; i++) 
   { 
       // Adding the previous 2 terms to make the 3rd term 
+      f[0] = f[1];
+      f[1] = f[2];
       f[i] = f[i-1] + f[i-2];
```
kvedala, line 30: will this line be correct anymore?
`f` was modified to an array of length 3 and hence, this loop will fail for `i>3`.

PR #524, opened 2020-04-20
Numerical methods

kvedala:
# Whats new
## Numerical Algorithms
1. [Durand Kerner method](https://en.wikipedia.org/wiki/Durand–Kerner_method) to compute real and complex roots of any polynomial with real coefficients.
2. [QR decomposition](https://en.wikipedia.org/wiki/QR_decomposition) for matrix with real elements.
3. Compute real Eigen values of a square matrix using QR-decomposition.

## Sorting Algorithms
1. Added another implementation for shell-sort.

## Online compile and run code
All the code can be executed and tested online: [using Google Colab Notebook](https://colab.research.google.com/gist/kvedala/27f1b0b6502af935f6917673ec43bcd7/plot-durand_kerner-log.ipynb)


kvedala:
@StepfenShawn

PR #532, opened 2020-05-18
Create AscendingPriorityQueue.c

shubhamsah:
Program for Ascending Priority Queue using Linked List

PR #538, opened 2020-05-23
Resolved PR #532

shubhamsah:


data_structures/linked_list/ascendingPriorityQueue.c
```diff
@@ -0,0 +1,169 @@
+/* Ascending priority queue using Linked List - Program to implement Ascending priority queue using Linked List */
+
```
deadshotsb, line 2: @shubhamsah Great work, kindly add a sample case of your program with the state of the queue in each operation, like m insert() followed by n delete() and exit() fundtions

sorting/MERGENR.C
```diff
@@ -0,0 +1,82 @@
+/* Program to demonstrate non recursive merge sort */
```
deadshotsb, line 1: Add a sample test case here also

data_structures/linked_list/ascendingPriorityQueue.c
shubhamsah: Thanks for the review, But could you specify an example so that I can get better understanding of what changes I need to make.

deadshotsb: @shubhamsah make some larger insert() and delete() calls to show the priority.
input        state of queue
1 
2                       2
1
3                        2->3
2                        3
3

Output
2

shubhamsah: Roger... Will do it

kvedala:
Then this PR can be closed

data_structures/linked_list/ascendingPriorityQueue.c
shubhamsah: Thank you for the review, 
I am still not able to understand the context of the review. Do you mean to add a function to test/run my code or should I comment out the output? Sorry to disturb you but even if you could give a previous file with a similar algorithm and the sample included I can learn and mold my file accordingly as I can't see a file that includes one such. And if there aren't any changes then I guess we are good to merge.

deadshotsb: My fault there is a indentation problem, just comment out the output . I have shown some, but use a larger number of insert().

shubhamsah: Yes I got it... but there's an error in your test case since this is an ascending priority queue so the smallest number in the queue would be removed
so the final output would be 3

sorting/MERGENR.C
shubhamsah: This merge sort is implemented using a nonrecursive approach. Adding a sample test case.

kvedala:
Can you please also include documentation to the code. This repository is meant for educational purposes so that others can read your code and learn from it.

shubhamsah:
Okay I will do It. Thanks for mentioning. Is there any changes in the code??

data_structures/linked_list/ascendingPriorityQueue.c
```diff
@@ -0,0 +1,271 @@
...
+#include  <stdio.h>
+#include  <stdlib.h>  
+#define  NULL  0
```
kvedala, line 5: this is a C++ syntax.
In C, NULL can be defined as:
```C
#define NULL ((void*)0)
```
Or it can be avoided altogether

data_structures/linked_list/ascendingPriorityQueue.c
```diff
@@ -0,0 +1,271 @@
...
+
+int  removes()
+{
+  int  min ;
+  struct  node  *follow , *follow1 , *p , *p1 ;
+
+  if(empty())
+  {
+    printf("\nQueue Underflow. Unable to remove.") ;
+    getch() ;
```
kvedala, line 61: `getch()` is a non-standard function and should be removed

kvedala:
Filenames should be all small letters without spaces including the suffix `.c`.

data_structures/linked_list/ascendingPriorityQueue.c
shubhamsah: Done.

data_structures/linked_list/ascendingPriorityQueue.c
shubhamsah: Done

shubhamsah:
All the said changes were done! Thank you.

kvedala:
Documentation was not to the standard as requested in the comments above. The documentation should be such that a lay person would be able to read and learn from your code. This is an educational repository. Hence, documentation should include at least the following:
1. brief file description
2. inline comments to understand the reasoning what the particular  line statement does
3. comments that describe the function, structure, etc.
Please [see this example](https://www.star.bnl.gov/public/comp/sofi/doxygen/docblocks.html)

PR #539, opened 2020-05-26
Automatic documentation generation using Doxygen and GitHub hooks

kvedala:
Documentation viewable at: https://kvedala.github.io/C/
Documentation updated automatically on push to master branch


kvedala:
wrong pull request. Was creating to a personal fork.

Nonetheless, @cclauss this could be a good way to keep track of files and the algorithms themselves instead of Directory.md which, I don't see any particular use for.

PR #547, opened 2020-06-17
added taylor series

deltanode:
Program to implement sum of Taylor Series(e^x)

### DIFF
--- a/misc/taylor_series.c
+++ b/misc/taylor_series.c
@@ -1,0 +1,56 @@
+/**
+ * C program to implement sum of Taylor Series(e^x)
+ * e^x (upto n-terms) = 1 + x^1/1! + x^2/2! + x^3/3! + ....... + x^n/n!
+ */
+
+#include <stdio.h>
+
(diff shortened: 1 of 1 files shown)


kvedala:
@yogesh7132 This is a specific implementation of approximation for `e^x` using Taylor series expansion unto `n` terms. Though good, it has some fundamental errors:
1. the factorial computation uses `double` precision arithmetics and hence already suffers from overflows and rounding off-errors because factorial is supposed to be an integer. To check, try to print the value of fact after 10 terms.
2. The implementation limits should be clearly specified and thus show how the code and the algorithm implementation are valuable.
3. please update the code with documentation and standards per the repo. See for example: [this code from repo](https://github.com/TheAlgorithms/C/blob/master/numerical_methods/ode_forward_euler.c) with documentation so that the code is self-explanatory and that the auto-tools can update the [documentation website](https://thealgorithms.github.io/C/d4/d07/ode__forward__euler_8c.html). Dont have to necessarily provide with graphs, but at least the documentation.

PR #548, opened 2020-06-20
Create b+tree.c

meltingOcean:
It is a code for b+tree. I have create insert operation and function that print whole of b+tree on tree shape.


data_structures/btree/b+tree.c
```diff
@@ -0,0 +1,381 @@
+#include<stdio.h>
+#include<stdlib.h>
```
Panquesito7, line 2:
```suggestion
#include <cstdio>
#include <cstdlib>
```

PR #573, opened 2020-07-19
bubble sort optimization

shellhub:
<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C/pull/573"></a>

### DIFF
--- a/sorting/bubble_sort.c
+++ b/sorting/bubble_sort.c
@@ -1,20 +1,33 @@
-// sorting of array list using bubble sort
+/**
+ * @file
+ * @brief [Bubble sort](https://en.wikipedia.org/wiki/Bubble_sort) algorithm
+ * implementation
+ */
+#include <assert.h>
+#include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <time.h>
 
-/*Displays the array, passed to this method*/
-void display(int *arr, int n)
+/**
+ * Display elements of array
+ * @param arr array to be display
+ * @param n length of array
+ */
+void display(const int *arr, int n)
 {
-    int i;
-    for (i = 0; i < n; i++)
+    for (int i = 0; i < n; i++)
     {
         printf("%d ", arr[i]);
     }
-
     printf("\n");
 }
 
-/*Swap function to swap two values*/
+/**
+ * Swap two values by using pointer
+ * @param first first pointer of first number
+ * @param second second pointer of second number
+ */
 void swap(int *first, int *second)
 {
     int temp = *first;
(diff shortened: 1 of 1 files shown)


sorting/bubble_sort.c
```diff
@@ -1,55 +1,73 @@
-// sorting of array list using bubble sort
+/**
+ * @file
+ * @brief [Bubble sort algorithm
```
kvedala, line 3: restructure the description as:
```
 * @brief [Bubble sort](https://en.wikipedia.org/wiki/Bubble_sort) algorithm implementation
```

sorting/bubble_sort.c
```diff
@@ -1,55 +1,73 @@
...
 #include <stdio.h>
 #include <stdlib.h>
 
-/*Displays the array, passed to this method*/
+/**
+ * Display elements of array
+ * @param arr array to be display
+ * @param n length of array
+ */
 void display(int *arr, int n)
```
kvedala, line 15:
```suggestion
void display(const int *arr, int n)
```

sorting/bubble_sort.c
```diff
@@ -1,55 +1,73 @@
...
     int temp = *first;
     *first = *second;
     *second = temp;
 }
 
-/*This is where the sorting of the array takes place
- arr[] --- Array to be sorted
- size --- Array Size
+/**
+ * Bubble sort algorithm implements
```
kvedala, line 37:
```suggestion
 * Bubble sort algorithm implementation
```

sorting/bubble_sort.c
```diff
@@ -1,55 +1,73 @@
...
+ * @param arr array to be sorted
+ * @param size size of array
  */
 void bubbleSort(int *arr, int size)
 {
     for (int i = 0; i < size - 1; i++)
     {
```
kvedala, line 44:
```suggestion
    { /* for each array index */
```

sorting/bubble_sort.c
```diff
@@ -1,55 +1,73 @@
...
+        bool swapped = false;
```
kvedala, line 45:
```suggestion
        bool swapped = false; /* flag to check if any changes had to be made */
        /* perform iterations until no more changes were made or outer loop
            executed for all array indices */
```

sorting/bubble_sort.c
```diff
@@ -1,55 +1,73 @@
...
         for (int j = 0; j < size - 1 - i; j++)
         {
```
kvedala, line 47:
```suggestion
        { /* for each element in the array */
```

sorting/bubble_sort.c
```diff
@@ -1,55 +1,73 @@
...
             if (arr[j] > arr[j + 1])
             {
```
kvedala, line 49:
```suggestion
            {/* if the order of successive elements needs update */
```

sorting/bubble_sort.c
```diff
@@ -1,55 +1,73 @@
...
                 swap(&arr[j], &arr[j + 1]);
+                swapped = true;
```
kvedala, line 51:
```suggestion
                swapped = true; /* set flag */
```

sorting/bubble_sort.c
```diff
@@ -1,55 +1,73 @@
...
             }
         }
+        if (!swapped)
+        {
+            break;
```
kvedala, line 56:
```suggestion
            /* since no more updates we made, the array is already sorted 
                this is an optimization for early termination */
            break;
```

sorting/bubble_sort.c
```diff
@@ -1,55 +1,73 @@
...
+        }
     }
 }
```
kvedala, line 60: Could you please add a test case with arrays of random numbers?

shellhub: @kvedala please review again. thx

kvedala:
The CI error can be fixed by using `calloc` instead of `malloc`.
Also, from next time, please use the provided Description template and not delete it. Thanks :)

shellhub:
> The CI error can be fixed by using `calloc` instead of `malloc`.
> Also, from next time, please use the provided Description template and not delete it. Thanks :)

Nice. error fixed. Please review again. Thanks :)

kvedala:
please implement `test()` function. I understand it was missing in the original implementation, but adding it now would be ideal. Thanks.

shellhub:
> please implement `test()` function. I understand it was missing in the original implementation, but adding it now would be ideal. Thanks.

``` c
/**
 * Test function
 */
void test()
{
    const int size = 10;
    int arr[size];

    /* generate size random numbers from 0 to 100 */
    for (int i = 0; i < size; i++)
    {
        arr[i] = rand() % 100;
    }
    bubbleSort(arr, size);
    for (int i = 0; i < size - 1; ++i)
    {
        assert(arr[i] <= arr[i + 1]);
    }
}
```
this test() function will be ok ?

Issue #729, opened 2020-10-29
[BUG] Unhandled Exception in Tic-tack-toe-game


devraj4522:
Tick-Tack-Toe game is not working for exception part
## Description

Else part on line 247 takes you to infinite loop on getting any char value as input for position.

## Expected Behavior

It should give the option to enter value again

## Actual Behavior

Infite Loop with print value - Invalid value

123mansha:
hii want to work on this issue please tell me the error and assign it to me

hero4127:
Hi want to work with this issue please assign it to as I have much experience in C program and I have excellent logic developed.

Panquesito7:
Please don't ask to be assigned. See the [contributing guidelines](https://github.com/TheAlgorithms/C-Plus-Plus/blob/master/CONTRIBUTING.md#contributor).
If you want to help with this, submit a PR and maintainers will review it.

Thank you. 🙂

kana800:
@Panquesito7  I fixed the bugs 😄

north8side:
Good work

north8side:
C++ is not the best for it's function needs more

PR #778, opened 2020-12-16
Update tic_tac_toe.c

driti924:
Successful exceptional handling

#### Description of Change

Issue: #729

Fixes / Changes

#line no 16
[+] #include <string.h>

First, I changed type of 'n' from int to char, so that if player enters any character other than a number it still gets consumed and does not remain in input buffer. As this input buffer effects the scanning of next number(i.e 'l' here) and doesn't scan player's choice to play again. 
#line no 41
[-] int n = 0;
[+] char n = '0';

Second, to avoid infinite loop on getting any character on line no.247 , I have added a new function 'TakeInput'. Thus a player can enter any character/s or number(not in range of 1-9) without entering infinite loop. This time a message will prompt them to give valid input.
Also note, my code is always using TakeInput function to input players choice to avoid the same problem of unused input buffer.
#line no 21
[+] static int TakeInput();   // used to take input from player and handle exceptions accordingly

#line no 85
[+] int TakeInput(){

    // Input player's choice of placing 'x' or 'o' in variable x 
    char x[100];
    fgets(x, 100, stdin);

    // to check the length of input!
    int i;
    for (i = 0; x[i] != '\0'; i++);   

    // if length of input=1 (i.e i=2) and input value is a digit then its a valid input.
    if(i==2 && x[0]-'0'>=1 && x[0]-'0'<=9  ){   
        int m=x[0]-'0';
        return m;
    }
    else{
        printf("(Enter a valid number between [1 - 9]) ");
        return TakeInput();
    }
}

Other changer are made accordingly. Like 
#line no 54
[-] scanf("%d", &n);
[+] scanf(" %c", &n);

#line no  59
[-] case 1:
[+] case '1':

#line no  62
[-] case 2:
[+] case '2':

For better visibility
#line no 66
 [-] printf("THANK YOU and EXIT!");
 [+] printf("THANK YOU and EXIT!\n");

# lines 105 and 176
[-] scanf("%d",&m)
[+] m=TakeInput() 

#line no 182
[-] scanf("%d", &e1);
[+] e1=TakeInput();

#line no 242 and 252
[-] scanf("%d", &n1);
[+] n1=TakeInput();

For better visibility
line no 272
[-] printf("\n Computer placed at %d position\n", e + 1);
[+] printf("\nComputer placed at %d position\n", e + 1);

#line no 301 and 311
[-] scanf("%d", &n1);
[+] n1=TakeInput();

### DIFF
--- a/games/tic_tac_toe.c
+++ b/games/tic_tac_toe.c
@@ -13,10 +13,12 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <time.h>
+#include <string.h>
 
 // Functions Declarations
 static void singlemode();
 static void doublemode();
+static int TakeInput();   // used to take input from player and handle exceptions accordingly
 static void placex(int);  // used for placing position of X by the 1st player
 static void place();      // used by the computer to place O
 static void placey(int);  // used in Double Player mode by the 2nd player to
@@ -38,7 +40,7 @@ int main()
     int l = 0;
     do
     {
-        int n = 0;
+        char n = '0';
 
         // filling the table with multiple asterisks
         for (int i = 0; i < 9; i++) game_table[i] = '*';
@@ -51,19 +53,19 @@ int main()
         printf("***********2. YOU vs PLAYER ***********\n");
         printf("***********3.EXIT *********************\n");
         printf("Enter your choice : ");
-        scanf("%d", &n);
+        scanf(" %c", &n);
 
         switch (n)  // switch case to select between single player mode or
                     // double player mode
         {
-        case 1:
+        case '1':
             singlemode();
             break;
-        case 2:
+        case '2':
             doublemode();
             break;
         default:
-            printf("THANK YOU and EXIT!");
+            printf("THANK YOU and EXIT!\n");
         }
 
         printf("Next game ? : ");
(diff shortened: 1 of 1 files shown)


driti924:
Hey this is Driti. This is my first opensource contribution.
One check is failing, I dont understand how to get rid of it as of now. I will try to resolve it.
Any feedback will be appreciated.

games/tic_tac_toe.c
```diff
@@ -80,6 +82,27 @@ int main()
...
  */
+int TakeInput(){
+    
+    // Input player's choice of placing 'x' or 'o' in variable x 
+    char x[100];
+    fgets(x, 100, stdin);
+
+    // to check the length of input!
+    int i;
+    for (i = 0; x[i] != '\0'; i++);
```
kvedala, line 93: in order to ensure this is valid, should the memory block of `x` be initialized to `\0` as:
```c
memset(x, 100, '\0');
```

games/tic_tac_toe.c
```diff
@@ -80,6 +82,27 @@ int main()
...
+        int m=x[0]-'0';
+        return m;
+    }
+    else{
+        printf("(Enter a valid number between [1 - 9]) ");
+        return TakeInput();
+    }
+}
+
 void singlemode()
```
kvedala, line 106: the documentation for this function was initially from line 78 to 82 and is misplaced.

games/tic_tac_toe.c
```diff
@@ -80,6 +82,27 @@ int main()
  *
  * @returns None
  */
+int TakeInput(){
```
kvedala, line 85: please update function documentation

kvedala (changes requested):
Good bug fix. But the updated code is not formatted. Please enable GitHub actions in your fork and it will clean up the code for you. 

The check that is failing will be resolved once we merge #796

PR #886, opened 2021-10-15
Algorithm to Split linked list into even and odd integers linked list.

the-suman:
Added an algorithm in C for splitting the given linked list into even and odd integer linked list.

As for example:-
Suppose my given linked list is: 12 45 30 7 10000 789
Then the result will be:
The even integer linked list is: 12 30 10000
The odd integer linked list is: 45 7 789

<a href="https://gitpod.io/#https://github.com/TheAlgorithms/C/pull/886"></a>


the-suman:
I am closing this request because it is not fulfilling your requirement.

PR #1031, opened 2022-10-10
Update to binary_insertion_sort.c

sarthaktanwani:
Pre-increment is supposed to be more efficient

#### Checklist

Notes:

### DIFF
--- a/sorting/binary_insertion_sort.c
+++ b/sorting/binary_insertion_sort.c
@@ -61,7 +61,7 @@ int main(int argc, const char *argv[])
     printf("Enter the elements of the array\n");
     int i;
     int *arr = (int *)malloc(n * sizeof(int));
-    for (i = 0; i < n; i++)
+    for (i = 0; i < n; ++i)
     {
         scanf("%d", &arr[i]);
     }
--- a/sorting/binary_insertion_sort.c
+++ b/sorting/binary_insertion_sort.c
@@ -3,6 +3,9 @@
  * inserting the selected item at each iteration. */
 #include <stdio.h>
 #include <stdlib.h>
+#include <assert.h>
+
+#define ARR_SZ_CAP	1024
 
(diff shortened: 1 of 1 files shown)


sorting/binary_insertion_sort.c
```diff
@@ -3,6 +3,9 @@
  * inserting the selected item at each iteration. */
 #include <stdio.h>
 #include <stdlib.h>
+#include <assert.h>
+
+#define ARR_SZ_CAP	1024
```
Panquesito7, line 8: Could you explain the use of this variable, please?

sarthaktanwani: The size of the input arrays is taken by the user and memory is directly allocated depending on that size. This is usually not safe as the user might set a very large value and without any sanitation of input, the program would try to allocate a very large memory and might have unexpected behavior. 

Assert makes sure that the input size is less than the max capacity, otherwise safely exits the program,

sarthaktanwani: **This is commit where I added it and the reason**

[Safely allocation a fixed max size of array as opposed to directly taking size as input which may be unsafe](https://github.com/TheAlgorithms/C/pull/1031/commits/a97cd20a8240eaf8bbb84dfc6416f95367377871) 
a97cd20

sorting/binary_insertion_sort.c
```diff
@@ -58,10 +61,12 @@ int main(int argc, const char *argv[])
     printf("Enter size of array:\n");
     scanf("%d", &n);  // E.g. 8
 
+    assert( n < ARR_SZ_CAP);
```
Panquesito7, line 64:
```suggestion
    assert(n < ARR_SZ_CAP);  // This is to ensure the program doesn't collapse
                            // with a very large value given by the user.
```

PR #1117, opened 2022-10-28
Added 8051_HAL

icebergelectronics:
8051_HAL is a set of C header files used to abstract the user from knowing the 8051 hardware architecture .  By importing these headers into your project hardware implementation of C on 8051 can be done with great ease and various c algorithms can be implemented on it


icebergelectronics:
There is a pdf file in this which is required to understand the working of the header functions

danghai:
Is it useful algorithm?

cclauss:
Given that this processor is from 1980, there is probably not much use for this code.
* https://en.wikipedia.org/wiki/Intel_8051

icebergelectronics:
8051 can be used for low cost electronics like toys . Also it can be used for educational purposes.😄

kvedala:
these are very specific Intel 8051 instructions and do not belong here. 
It would be a good idea to create a hardware algorithms project and work on it there.

PR #1176, opened 2022-12-09
Fix Awesome Workflow

tjgurwara99:
#### Description of Change

Currently the awesome workflow's are failing for all of the PR. The reason for it is that the ubuntu repository has removed the package that we used to install `clang-format-10` and `clang-tidy-10` and it has been replaced by other packages. I propose that these packages shouldn't be linked to a static version in order to make sure that this kind of situation doesn't occur again.

PS: I had also renamed the Codeql file and made some changes, I don't think there is any difference to the one that was already commit to it except that runs-on being a array of different environments to build the codeql analysis for. If it turns out that you don't require this change, it would be easy for me to remove that from this branch, just let me know.

#### Checklist


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,74 @@
...
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "master" ]
+  schedule:
+    - cron: '23 12 * * 4'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: [ubuntu-latest, windows-latest, macos-latest]
```
Panquesito7, line 26: This won't work properly. I believe we need to use a matrix.
```suggestion
    runs-on: ${{ matrix.os }}
```

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,74 @@
...
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'cpp' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
+        # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
```
Panquesito7, line 37:
```suggestion
        # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
        os: [windows-latest, ubuntu-latest, macOS-latest]
```

Panquesito7 (changes requested):

.github/workflows/codeql.yml
tjgurwara99: I think this won't work because the scope of `os` is above the definition of the `matrix.os`. I've just used `ubuntu-latest` for now. We can change it if something changes 😄

.github/workflows/codeql.yml
Panquesito7: > Are you sure?

I believe so. I've done it in other repositories and it doesn't work. When I use `matrix.os`, it works.

Panquesito7 (approved):
Looks good. Thanks! 🚀
The other stuff can be addressed in another PR. This is something that must be fixed ASAP.

.github/workflows/codeql.yml
tjgurwara99: Oh I think you're right. I will see what I can do 😄

PR #1184, opened 2022-12-15
add leetcode Longest Chunked Palindrome Decomposition

alexpantyukhin:
#### Description of Change

Add a guide for new LeetCode solutions.

#### Checklist

Notes:


leetcode/src/1147.c
```diff
@@ -0,0 +1,49 @@
+#define max(a,b) (((a)>(b))?(a):(b))
+
+bool substringsAreEqual(char* text, int firstIndex, int secondIndex, int length){
```
Panquesito7, line 3: We could probably rename the function's name here. What do you think?
```suggestion
bool equalSubstrings(char* text, int firstIndex, int secondIndex, int length){
```

alexpantyukhin: Yes, sure! good idea

Panquesito7 (approved):
Thanks! 🚀

leetcode/src/1147.c
```diff
@@ -0,0 +1,49 @@
...
+    return dp[index];
+}
+
+int longestDecompositionDp(char* text, int textLen, int index, int* dp){
+    char ch = text[index];
+    int result = 1;
+    
+    for (int i = 0; i < (textLen - 2 * index) / 2; i++){
+        if (ch == text[textLen - 1 - index - i] 
+            && equalSubstrings(text, index, textLen - 1 - index - i, i + 1)){
```
tjgurwara99, line 31: Again this is something general, but whenever you come across long conditional statements like these, it is better to name the conditions for easy readability. The way to do this is create a function with the name of the condition and the inputs it takes. The arguments shouldn't do any calculations by themselves but arguably this is fine for now 😄.

Always know that code is read much more often than written so it may be easier to write things like this when you have the context since its fresh in mind but when you come back to it after a while, even you who's written it will have to figure out why you wrote something.

tjgurwara99 (approved):

PR #1219, opened 2023-02-21
Add Shunting Yard Algorithm

CascadingCascade:
#### Description of Change
Add an implementation of the shunting yard algorithm

#### Checklist

Notes:


misc/shunting_yard.c
```diff
@@ -0,0 +1,222 @@
...
+    const unsigned int inputLength = strlen(input);
+    char* operatorStack = (char*) malloc(sizeof(char) * inputLength);
+
+    // This pointer points at where we should insert the next element,
+    // Hence stackPointer - 1 is used when accessing elements
+    unsigned int stackPointer = 0;
+
+    // We will parse the input with strtok(),
+    // Since strtok() is destructive, we make a copy of the input to preserve the original string
+    char* str = malloc(sizeof(char) * inputLength);
```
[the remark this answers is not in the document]
CascadingCascade: Oh damn, how come my linter didn't catch this, and it passed the tests too:sweat_smile:

misc/shunting_yard.c
```diff
@@ -0,0 +1,225 @@
...
+/**
+ * @brief Helper function that returns each operator's associativity
+ * @param operator the operator to be queried
+ * @returns '1' if the operator is left associative
+ * @returns '0' if the operator is right associative
+ */
+int getAssociativity(char operator) {
+    switch (operator) {
+        case '^': {
+            return 0;
```
Panquesito7, line 51: Shouldn't you add `break` for these statements?

CascadingCascade: I thought `break` aren't necessary since they `return` instead?

misc/shunting_yard.c
```diff
@@ -0,0 +1,225 @@
...
+        switch (token[0]) {
+            // If it's a left parenthesis, push it to the operator stack for later matching
+            case '(': {
+                operatorStack[stackPointer++] = token[0];
+                break;
+            }
+
+            // If it's a right parenthesis, search for a left parenthesis to match it
+            case ')': {
+                while (operatorStack[stackPointer - 1] != '(') {
```
Panquesito7, line 110: For some reason, the `clang-tidy` checker on Gitpod is reporting a warning here. 🤔
Weird that there's no warning in the GitHub CI.

Panquesito7 (approved):
Otherwise, it's looking awesome. Thanks! 🚀

misc/shunting_yard.c
CascadingCascade: That's certainly very interesting, my local `clang-tidy` reported nothing (Otherwise I would have fixed it before upload) Lemme guess, it claims `loop variable not updated`? (or something like that, I don't remember the exact wording) It's probably a glitch or something, my `clang-tidy` just won't stop screaming when I tried to make the algorithm not output extra space after the final operator because `condition always true` obviously it's not always true because `stackPointer` is decreased in the middle of the loop, but it insists anyway, and I had to give up. my guess is that `strncat(output,&operatorStack[(stackPointer--) - 1],1);` confused it for whatever reason:thinking:

CascadingCascade: I've opened this in Gitpod myself and it claims `the left operand is a garbage value`, it's actually not without merit though, if someone hypothetically inputs something blatantly wrong like ") 3 + 4 (" there can potentially be a segfault. I've tried adding a guard statement, but somehow it continues to warn me anyway, could use some advice before I commit this change

Panquesito7 (approved):

Panquesito7 (approved):

alexpantyukhin (approved):

PR #1222, opened 2023-02-23
feat: remove nth node from end of list (leetcode #19)

mindaugl:
#### Checklist

- [ ] Added description of change

Notes:


leetcode/src/19.c
```diff
@@ -0,0 +1,32 @@
...
+        sz++;
+    }
+    if (n == sz) {
+        p_free = head;
+        p = head->next;
+        free(p_free);
+        return p;
+    }
+    for (i = 0, p = head; i < sz - n - 1; i++, p = p -> next)
+    ;
```
CascadingCascade, line 23: This `for` with an empty body surely looks scary, while it does it's job just fine, you should probably put `p = p -> next` in the loop body, for readability's sake if nothing else. also, if `i` is only used here as the loop variable, why not declare it here instead?
```suggestion
    for (int i = 0, p = head; i < sz - n - 1; i++) {
        p = p -> next;
    }
```

leetcode/src/19.c
```diff
@@ -0,0 +1,32 @@
...
+ * };
+ */
+
+struct ListNode *removeNthFromEnd(struct ListNode *head, int n) {
+    struct ListNode *p_free, *p = head;
+    int i, sz = 0;
+    while (p != NULL) {
+        p = p->next;
+        sz++;
+    }
```
CascadingCascade, line 15: It would be more efficient if you use memoization, I checked the problem description, an array of 30 pointers should be plenty enough. That way you get to solve the follow up objective of only one pass

CascadingCascade (commented):
This code should work, but I've came up with some enhancement ideas

leetcode/src/19.c
mindaugl: You cannot declare variable inside for loop fields in C (unlike C++). For the pointer increment inside for loop, it is more of a style taste, happy to change if more reviewers feel the same way as you do, thanks

leetcode/src/19.c
mindaugl: Thanks for the comment, I agree that it is possible to solve the problem with memoization and I considered going this route, but decided to make algorithm more generic (allowing more than 30 nodes). Memoization approach would require to trade space for speed and I think would be an alternative solution all together, so happy to implement it at the later stage (or someone else will do it).

leetcode/src/19.c
CascadingCascade: Oh, I have no intention to force anything on you, sorry if my comment came out as more blunt than I had intended it to be, but I'm pretty sure `for(int i = 0; i < n; i++)` is a thing in C

mindaugl: Sorry, yes you are right, it was introduced in C99 (https://stackoverflow.com/questions/1287863/c-for-loop-int-initial-declaration), I was thinking about the earlier standards :) Nevertheless, if putting "int" inside the for initialization field, would need to move "p = head" outside the for() initialization field, thanks

leetcode/src/19.c
```diff
@@ -0,0 +1,32 @@
...
+    if (n == sz) {
```
alexpantyukhin, line 16: What if we avooid this check with the following version of the code:
```
    struct ListNode sentry;
    sentry.next = head;

    struct ListNode *p_free, *p = &sentry;
    int i, sz = 0;

    while (p != NULL) {
        p = p->next;
        sz++;
    }

    p = &sentry;
    for (int i = 0; i < sz - n - 1; i++) {
        p = p->next;
    }
    
    struct ListNode *p_next_next = p->next->next;
    // we check only there if we need to free the resource.
    if (p_next_next == NULL) {
        free(p_next_next);
    }

    p->next = p_next_next;
    return sentry.next;
```
Would it be good? what do you think?

CascadingCascade: I second this idea, the code is definitely cleaner this way. I should note that @alexpantyukhin's example code does contains some problems (duplicate `i` declaration, for example), but the idea's good

mindaugl: I do not fully follow your suggestion, but if I understand correctly, you propose to introduce new entry node, so that there is no need to check if n == 1. After introducing entry node, I think the code should be as below. Need to call free() irrespective of value of n.

```
struct ListNode *removeNthFromEnd(struct ListNode *head, int n) {
    struct ListNode entry, *p_free, *p = head;
    entry.next = head;
    int sz = 0;
    while (p != NULL) {
        p = p->next;
        sz++;
    }
    p = &entry;
    for (int i = 0; i < sz - n; i++, p = p -> next)
    ;
    p_free = p->next;
    if (n != 1) {
        p->next = p->next->next;
    } else {
        p->next = NULL;
    }
    free(p_free);
    return entry.next;
}
```

mindaugl: Pushed the commit for this change

alexpantyukhin (approved):

Panquesito7 (approved):
Thanks! 🚀

PR #1221, opened 2023-02-22
feat: add Letter combinations of phone book problem (#17)

mindaugl:
#### Checklist

- [ ] Added description of change
- [ ] Added file name matches [File name guidelines](https://github.com/TheAlgorithms/C/blob/master/CONTRIBUTING.md#File-Name-guidelines)
- [ ] Added tests and example, test must pass
- [ ] Relevant documentation/comments is changed or added
- [ ] PR title follows semantic [commit guidelines](https://github.com/TheAlgorithms/C/blob/master/CONTRIBUTING.md#Commit-Guidelines)
- [ ] Search previous suggestions before making a new one, as yours may be a duplicate.
- [ ] I acknowledge that all my contributions will be made under the project's license.

Notes:

PR #1226, opened 2023-02-26
feat: solution for Merge k Sorted Lists problem (#23)

mindaugl:
#### Checklist

- [ ] Added description of change

Notes:


leetcode/src/23.c
```diff
@@ -0,0 +1,67 @@
+/**
+ * Definition for singly-linked list.
+ * struct ListNode {
+ *     int val;
+ *     struct ListNode *next;
+ * };
+ * 
+ * Algorithm merges pairwaise lists by divide on conquer.
```
CascadingCascade, line 8:
```suggestion
 * Algorithm merges pairwise lists by divide and conquer.
```

leetcode/src/23.c
```diff
@@ -0,0 +1,67 @@
...
+    return head;
+}
+
+struct ListNode *mergeKLists(struct ListNode** lists, int lists_size) {
+    if (lists_size == 0) {
+        return NULL;
+    }
+    struct ListNode **ml;
+    int i, is_odd, sz = lists_size;
+    ml = malloc(sizeof(struct ListNode*) * lists_size);
```
CascadingCascade, line 50: Memory leak? You returned `ml[0]` directly, it's probably better to copy the pointer, free `ml` , then return the copy, alternatively, maybe declare `struct ListNode* ml2[lists_size];` instead?

CascadingCascade (commented):
Hey, you should sync your fork and enable actions, all your PRs are `branch out-of-date`, #1222 even have a merge conflict.

leetcode/src/23.c
mindaugl: Just fixed the typos. Regarding "enable actions", I think it is enabled, at least it was clicked when I created PR. Also fixed merge conflict for the #1222 PR.

leetcode/src/23.c
mindaugl: Yes, fixed the memory leak, check the updated code, thanks

leetcode/src/23.c
CascadingCascade: Oh, I am sure actions are enabled at the repository level, not when you open PRs, I didn't see any workflow run [here](https://github.com/mindaugl/C/actions). Also #1221 is still out-of-date, you might also want to sync your fork's master branch and delete branches whose PR has already been merged

mindaugl: I just enabled workflow runs and merged #1221

Panquesito7 (approved):
LGTM. Thanks! 🚀

leetcode/src/23.c
```diff
@@ -0,0 +1,69 @@
...
+    if (lists_size == 0) {
+        return NULL;
+    }
+    struct ListNode **ml, *head;
+    int i, is_odd, sz = lists_size;
+    ml = malloc(sizeof(struct ListNode*) * lists_size);
+    for (i = 0; i < lists_size; i++) {
+        ml[i] = lists[i];
+    }
+    while(sz > 1) {
```
alexpantyukhin, line 54: I think you can use more simple approach:

the function `mergeTwo(target, source)` - merges target and source into the target in the sorted order.
the function `mergeKLists` :
 - creates a result head.
 - uses `for (lists_size)`: 
 -   each loop uses mergeTwo(head, curr_list).

I believe this case you will have the Space complexity: O(1), Runtime: (Sum(listsizes)).

mindaugl: Yes, it will use O(1) space complexity, but will use runtime complexity of O(k^2 * n), where k is the numbers of lists, and n is the average list size - such run complexity is lower than the current implementation, which is O(k * n * log(k)).

alexpantyukhin:
This is good task. But I believe the code could be made cleaner. 
This is my example of the python approach:
```python
    def mergeKLists(self, lists: List[ListNode]) -> ListNode:
        head = ListNode(-1, None)
        cursor = head
        while (True):
            minimum = 1000000
            head_ro_remove = None
            head_ro_remove_index = None
            all_lists_are_empty = True
            
            for (index, hl) in enumerate(lists):
                if hl is not None:
                    all_lists_are_empty = False

                    if hl.val < minimum:
                        minimum = hl.val
                        head_ro_remove_index = index

            if all_lists_are_empty:
                break
               
            cursor.next = ListNode(lists[head_ro_remove_index].val, None)
            cursor = cursor.next
            lists[head_ro_remove_index] = lists[head_ro_remove_index].next
                
        return head.next
```

Here I use the `head` this is `sentry`, this avoid to make exta if's for example. Later I will back with the C variant.

CascadingCascade:
> This is good task. But I believe the code could be made cleaner. This is my example of the python approach:
> 
> ```python
>     def mergeKLists(self, lists: List[ListNode]) -> ListNode:
>         head = ListNode(-1, None)
>         cursor = head
>         while (True):
>             minimum = 1000000
>             head_ro_remove = None
>             head_ro_remove_index = None
>             all_lists_are_empty = True
>             
>             for (index, hl) in enumerate(lists):
>                 if hl is not None:
>                     all_lists_are_empty = False
> 
>                     if hl.val < minimum:
>                         minimum = hl.val
>                         head_ro_remove_index = index
> 
>             if all_lists_are_empty:
>                 break
>                
>             cursor.next = ListNode(lists[head_ro_remove_index].val, None)
>             cursor = cursor.next
>             lists[head_ro_remove_index] = lists[head_ro_remove_index].next
>                 
>         return head.next
> ```
> 
> Here I use the `head` this is `sentry`, this avoid to make exta if's for example. Later I will back with the C variant.

Got to agree this approach would make the code cleaner and more readable, but I am of the opinion that the original divide and conquer solution is more efficient. What should we prioritize?:thinking: Also I mixed in some (dynamic programming?) and cooked up this diabolical piece of untested code that may or may not work and may or may not be interesting:stuck_out_tongue_winking_eye::
```c
int cmpNode(struct ListNode* node1, struct ListNode* node2) {
    if (node1 == NULL) return 1;
    if (node1->val > node2->val) return 1;
    if (node1->val == node2->val) return 0;
    return -1;
}

struct ListNode *mergeKLists(struct ListNode** lists, int lists_size) {
    if (lists_size == 0) {
        return NULL;
    }
    const struct ListNode fictMax = {INT_MAX, NULL};
    struct ListNode* cursor = &fictMax;
    while (1) {
        struct ListNode *min = &fictMax, *secMin = &fictMax;
        int minDex = -1, secMinDex = -1;
        int endFlag = 1;

        for (int i = 0; i < lists_size; ++i) {
            if (lists[i] != NULL) {
                endFlag = 0;
                if (cmpNode(lists[i], min) <= 0) {
                    secMin = min;
                    secMinDex = minDex;
                    min = lists[i];
                    minDex = i;
                }
            }
        }

        if (minDex != - 1) {
            lists[minDex] = lists[minDex]->next;
        }
        if (secMinDex != -1) {
            lists[secMinDex] = lists[secMinDex]->next;
        }

        if (minDex != -1) {
            cursor->next = min;
            cursor = cursor->next;
        }
        if (secMinDex != -1) {
            cursor->next = secMin;
            cursor = cursor->next;
        }

        if (endFlag) break;
    }
    return fictMax.next;
}

```

PR #1228, opened 2023-03-03
feat: add LeetCode problem 61

ashikjhonson:
Rotate List

#### Checklist

Notes:


leetcode/src/61.c
```diff
@@ -0,0 +1,47 @@
...
+* Definition for singly-linked list.
+* struct ListNode {
+* int val;
+* struct ListNode *next;
+* };
+*/
+
+struct ListNode* rotateRight(struct ListNode* head, int k){
+struct ListNode *ptr = head, *prev =NULL;
+int count = 0;
```
Panquesito7, line 23: This code should be in a function to have it more structured, IMO.

leetcode/src/61.c
```diff
@@ -0,0 +1,47 @@
...
+        prev = ptr;
+        ptr = ptr->next;
+    }
+    ptr->next = head;
+    head = ptr;
+    prev->next = NULL;
+}
+return head;
+
+}
```
Panquesito7, line 47: Missing an endline. Without this endline, there could be issues with some IDEs and configurations.
```suggestion
}

```

Panquesito7 (changes requested):

Panquesito7 (dismissed):
Thanks! 🚀

leetcode/src/61.c
```diff
@@ -0,0 +1,48 @@
...
+    }
+    if(count!=0)
+        k = k%count;
+    while(k--) {
+        ptr = head;
+        if(ptr==NULL)
+            return NULL;
+        if(ptr->next==NULL)
+            return ptr;
+        while(ptr->next) {
```
alexpantyukhin, line 38: Are you sure that the Runtime complexity is O(n + k)? I see the outer loop is O(k) and the inner loop goes throught all the list so, If I'm right it's the Runtime O(n * k), no?

CascadingCascade: > Are you sure that the Runtime complexity is O(n + k)? I see the outer loop is O(k) and the inner loop goes throught all the list so, If I'm right it's the Runtime O(n * k), no?

I think you are right that it's not O(n + k), but I dissent that the outer loop is really O(k mod n), so O(n * k) would be the worst case time complexity (When k = n - 1).

CascadingCascade (commented):
I would stand by my earlier comment at the closed PR that your solution is not exactly an efficient one. If we keep track of three nodes: node at index `k % count` (pk), node at index `k % count - 1` (pk1) and list tail (tail), we can rotate the list by simply `tail.next = head; pk1.next = NULL; return pk;`. This way we get to solve the problem consistently in O(n) because everything except for the first loop only takes constant time.

PR #1232, opened 2023-03-09
Update fibonacci.c

niranjank2022:
#### Description of Change
Included some documentation and modified the base condition and return data-type of fib() function.

#### Checklist

Notes: 
I am a beginner. If I need to change or improve anything, kindly say. I would rectify it.


math/fibonacci.c
```diff
@@ -1,23 +1,40 @@
+/**
+ * @file fibonacci.c
+ * @brief Program to print the nth term of the Fibonacci series.
+ * @details Fibonacci series generally starts from 0 and 1. Every next term in
+ *the series is equal to the sum of the two preceding terms.
+ **/
+
 #include <stdio.h>
```
CascadingCascade, line 8:
```suggestion
#include <stdio.h> /// for IO operations
```

math/fibonacci.c
```diff
@@ -1,23 +1,40 @@
...
-int fib(int number)
-{
-    if (number == 1 || number == 2)
-        return 1;
-    else
-        return fib(number - 1) + fib(number - 2);
-}
+unsigned fib(int);
 
+/* Driver Code */
```
CascadingCascade, line 12:
```suggestion
/**
 * @brief Main function
 * @returns 0 on exit
 */
```

math/fibonacci.c
```diff
@@ -1,23 +1,40 @@
+/**
+ * @file fibonacci.c
```
CascadingCascade, line 2: No need to enter the file name manually, best let Doxygen detect it automatically
```suggestion
 * @file 
```

math/fibonacci.c
```diff
@@ -1,23 +1,40 @@
...
+ * 'unsigned' respectively (takes more than 20 seconds to print)
+ */
+unsigned fib(int number)
+{
+    if (number == 1)
+        return 0;
+    else if (number == 2)
+        return 1;
+    else
+        return fib(number - 1) + fib(number - 2);
```
CascadingCascade, line 39: There's no need for `else`, since the function returns in each case
```suggestion
    if (number == 1) {
        return 0;
    }
    if (number == 2) {
        return 1;
    }
    return fib(number - 1) + fib(number - 2);
```

math/fibonacci.c
```diff
@@ -1,23 +1,40 @@
...
-{
-    if (number == 1 || number == 2)
-        return 1;
-    else
-        return fib(number - 1) + fib(number - 2);
-}
+unsigned fib(int);
 
+/* Driver Code */
 int main()
```
CascadingCascade, line 13: This code was written a long time ago, since it looks like you are trying to improve it, you might want to add a `test()` to bring it up to the current standard

math/fibonacci.c
```diff
@@ -1,23 +1,40 @@
...
+
+/**
+ * @brief Determines the nth Fibonacci term
+ * @param number - n in "nth term" and it can't be negative
+ * @return nth term in unsigned type
+ * @warning Only till 47th and 48th fibonacci element can be stored in 'int' and
+ * 'unsigned' respectively (takes more than 20 seconds to print)
+ */
+unsigned fib(int number)
+{
```
CascadingCascade, line 33: Maybe you should make `number` `unsigned` too, or a check like this would be great
```suggestion
{
    if (number < 1) {
        fprintf(stderr,"Illegal Argument!");
        exit(EXIT_FAILURE);
    }
```

math/fibonacci.c
```diff
@@ -1,23 +1,40 @@
...
+    if (number == 1)
+        return 0;
+    else if (number == 2)
```
alexpantyukhin, line 36: Why do you change the value in 1st number value?

If that, what if to avoid `else`?
something like:
```c
if (number == 1)
    return 0;

if (number == 2)
    return 1;

return fib(number - 1) + fib(number - 2);
```

CascadingCascade: [Wikipedia](https://en.wikipedia.org/wiki/Fibonacci_number), the current definition starts the sequence from `0`

alexpantyukhin: I see there are many ways where to start from.

CascadingCascade: I suppose that's also true, 🤔

math/fibonacci.c
niranjank2022: I am not aware of what Doxygen is.

math/fibonacci.c
```diff
@@ -1,23 +1,40 @@
...
 
-// Fibonnacci function
-int fib(int number)
-{
-    if (number == 1 || number == 2)
-        return 1;
-    else
-        return fib(number - 1) + fib(number - 2);
-}
+unsigned fib(int);
```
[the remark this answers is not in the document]
niranjank2022: I am sorry for this. The books that I have been following were all using function declaration and definition. So, I am used to it. Is it something I need to change?

math/fibonacci.c
niranjank2022: Why this?

math/fibonacci.c
niranjank2022: Got it!

math/fibonacci.c
CascadingCascade: ...I see that you probably didn't read the contribution guidelines carefully, [here's the link](https://github.com/TheAlgorithms/C/blob/master/CONTRIBUTING.md)

math/fibonacci.c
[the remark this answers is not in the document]
CascadingCascade: Maybe, for large files it can be good for readability to put all the declarations and documentations at the file's beginning then write all the definitions, it's unnecessary to split them here, but it's not necessarily wrong either

math/fibonacci.c
CascadingCascade: It's what we use to generate the docs from the code, the contribution guidelines have more details, or you can check out their project website [here](https://www.doxygen.nl/index.html)

niranjank2022: Thanks for the help. I would check it.

math/fibonacci.c
CascadingCascade: It's some boilerplate docs for `main()` from the contribution guidelines, [link](https://github.com/TheAlgorithms/C/blob/master/CONTRIBUTING.md#typical-structure-of-a-program)

math/fibonacci.c
niranjank2022: May I close this PR and add another PR with the corrections mentioned here?

CascadingCascade: That's not necessary, if you haven't noticed, you should know that the entire PR will be squashed into a single commit after merge, just push your update to your fork.

math/fibonacci.c
```diff
@@ -1,23 +1,70 @@
...
+
+/**
+ * @brief self-test implementation
+ * @return void
+ */
+static void test()
+{
+    assert(fib(5));
+    assert(fib(2));
+    assert(fib(9));
```
CascadingCascade, line 49: That's not how assertions work, the `assert()` macro is used to test a condition, with the idea being that if the program is working properly, that condition will hold at that point in execution. If you input a function's return value directly, then it doesn't do anything except check if the return value is not `NULL`.
```suggestion
    assert(fib(5) == 3);
    assert(fib(2) == 1);
    assert(fib(9) == 21);
```

math/fibonacci.c
```diff
@@ -1,23 +1,70 @@
...
-    printf("Number: ");
-    scanf("%d", &number);
+    // Performing the test
+    test();
+    printf("Tests passed...\n");
 
-    printf("%d \n", fib(number));
+    // Getting n
+    printf("Enter n to find nth fibonacci element: ");
+    scanf("%d", &number);
```
CascadingCascade, line 66: What if the user inputted something other than a valid number? It's quite interesting to write input checks, and you should certainly consider adding some, but it's probably a bit too much, just ask me if you need assistance.😄 [And here's a great resource](https://sekrit.de/webdocs/c/beginners-guide-away-from-scanf.html)

niranjank2022: This is very much helpful!

niranjank2022: If characters other than numbers are entered, they are detected by the first base condition and the code exits.

CascadingCascade: > If characters other than numbers are entered, they are detected by the first base condition and the code exits.

That would be the case if you had initialized `number` to, say, `-1` and `scanf()` didn't parse anything. But no, since `number` is not initialized except by this `scanf()`, bad things can happen. I want challenge you to write a proper input check, it would be a great exercise.

math/fibonacci.c
```diff
@@ -56,14 +56,19 @@ static void test()
...
+    char buffer[8];
 
     // Performing the test
     test();
     printf("Tests passed...\n");
 
     // Getting n
     printf("Enter n to find nth fibonacci element: ");
-    scanf("%d", &number);
+    scanf("%2[0-9]s", buffer);
```
CascadingCascade, line 67: That `s` in `%2[0-9]s` is excessive, `%2[0-9]` already specifies 2 digits. Also consider `12z`. This `scanf()` would parse `12` and store it into buffer, even though the input as a whole is not a valid number. There's also no code to repeat the request for input in case of invalid input. Here's a useful [link](https://devdocs.io/c/io/fscanf).

math/fibonacci.c
```diff
@@ -56,14 +56,19 @@ static void test()
...
+
+    number = atoi(buffer);
```
CascadingCascade, line 69: `atoi()` is not a part of the C standard, use `strtol()` or `sscanf()` instead.

CascadingCascade (commented):
See, this is harder than what you probably expected. I can write some example code for you later if you need so, or maybe this exercise is going out of the scope of this PR and you would rather just stick to `scanf("%d",&number);`?

math/fibonacci.c
niranjank2022: What's the difference between keeping and not keeping 's' in the format specifier?

CascadingCascade: > What's the difference between keeping and not keeping 's' in the format specifier?

Like I've said, `%2[0-9]` already is a format specification, so the `s` becomes a character literal, `%2[0-9]s` matches 2 digits plus a single `s`, like `20s`

niranjank2022: What can I have as the buffer size?

niranjank2022 (commented):
is  the "else if" condition in line 61 necessary as it is covered in "if" condition at line 59

math/fibonacci.c
niranjank2022: How come it is not part of the C standard? It is present in <stdlib.h>, right?

math/fibonacci.c
```diff
@@ -1,23 +1,109 @@
...
+    do
+    {
+        fgets(buffer, 4, stdin);  // Inputs the value from user
+
+        errno = 0;  // Reseting the errno value
+        num = strtol(buffer, &endPtr,
+                     10);  // Attempts to convert the string to integer
+
+        if (strlen(buffer) > 3 || num > 48)
+            success = 0;  // Three digit numbers are not allowed
```
alexpantyukhin, line 60: why do we use `success` val instead of using `continue` (success = 0) and `break` instead of success = 1?

math/fibonacci.c
CascadingCascade: > How come it is not part of the C standard? It is present in <stdlib.h>, right?

That's another thing your teacher didn't told you. You most likely know that there's a lot of different C compilers out there(gcc, MSVC, clang, just to name a few), and in order for a compiler to be a C compiler, it only needs to include all features specified in the C standard, how these features are implemented and what additional features to include are entirely up the compiler's developer's discretion. `atoi()` is *not* a part of the ANSI standard, but it is a commonly included additional feature, hence it is present in *your copy* of the `stdlib.h`

math/fibonacci.c
```diff
@@ -38,6 +39,42 @@ unsigned int fib(int number)
...
+        else if (errno == ERANGE)
```
CascadingCascade, line 61: This check is redundant, since only up to three characters are read from input, it's impossible for an overflow to happen, not to mention there's already a `num > 48` check

math/fibonacci.c
```diff
@@ -38,6 +39,42 @@ unsigned int fib(int number)
...
+ * @return valid argument to the fibonacci function
+ */
+int getInput(void)
+{
+    int num, success = 0;
+    char buffer[4], *endPtr;
+
+    do
+    {
+        fgets(buffer, 4, stdin);  // Inputs the value from user
```
CascadingCascade, line 53: There's no prompt to reenter the number in case of invalid input

math/fibonacci.c
```diff
@@ -38,6 +39,42 @@ unsigned int fib(int number)
...
+        else if (endPtr == buffer)
+            success = 0;  // No character is entered
+        else if (*endPtr != '\0' && *endPtr != '\n')
+            success = 0;  // *endptr is neither end of string nor newline
+        else
+            success = 1;
+
+    } while ((!success));  // Repeat until valid
+                           // number is entered
+
```
CascadingCascade, line 73: Here, I cleaned this code up for you.:smile:
```suggestion
    int num;
    char buffer[4], *endPtr;

    while (1) { // Repeat until a valid number is entered
        printf("Please enter a valid number:");
        fgets(buffer, 4, stdin);  // Inputs the value from user

        num = strtol(buffer, &endPtr, 10);  // Attempts to convert the string to integer

        // Checking the input
        // Characters other than digits are included in the input
        if ( ( *endPtr != '\0' && *endPtr != '\n' ) ||
            // No characters are entered
            endPtr == buffer ||
            // The number is too large
            ( strlen(buffer) > 3 || num > 48 ) ) {

            continue;
        }

        break;
    }

```

CascadingCascade (commented):
As a beginner, you are doing great!:+1: There's still one issue: if the user inputs a large number like `3443434`, the program will read from this input repeatedly then output the 4th Fibonacci number. Can you fix this? 😉

CascadingCascade:
By the way, sync your fork.

math/fibonacci.c
niranjank2022: Then what will be the while condition

niranjank2022:
> As a beginner, you are doing great!👍 There's still one issue: if the user inputs a large number like `3443434`, the program will read from this input repeatedly then output the 4th Fibonacci number. Can you fix this? 😉

Won't this be addressed by increasing the buffer size?

math/fibonacci.c
niranjank2022: Can you please share any resources to check whether a function belongs to the C standard?

CascadingCascade:
> > As a beginner, you are doing great!👍 There's still one issue: if the user inputs a large number like `3443434`, the program will read from this input repeatedly then output the 4th Fibonacci number. Can you fix this? 😉
> 
> Won't this be addressed by increasing the buffer size?

Yes and no. No matter how large the buffer size is, it is still finite, and this will happen again if I input, say, a 300 digits number. A better method would be to consume and discard all remaining characters from the input stream with a loop after the first two characters have been consumed.

math/fibonacci.c
CascadingCascade: > Can you please share any resources to check whether a function belongs to the C standard?

Unfortunately I am not aware of any resources of this kind, but generally speaking, searching "[function name] best practice" should yield some useful results

niranjank2022: Thanks

niranjank2022 (commented):
Any modifications now?

CascadingCascade:
> Any modifications now?

You broke the no characters entered check, otherwise LGTM

niranjank2022:
> > Any modifications now?
> 
> You broke the no characters entered check, otherwise LGTM

I don't get it.

CascadingCascade:
> > > Any modifications now?
> > 
> > You broke the no characters entered check, otherwise LGTM
> 
> I don't get it.

Like I've said in the updated comment, now in these cases the program will get stuck unless you press `enter` again.

niranjank2022:
> > > > Any modifications now?
> > > 
> > > 
> > > You broke the no characters entered check, otherwise LGTM
> > 
> > 
> > I don't get it.
> 
> Like I've said in the updated comment, now in these cases the program will get stuck unless you press `enter` again.

I guess I resolved that now

math/fibonacci.c
```diff
@@ -1,23 +1,112 @@
...
+    while (1)
+    {  // Repeat until a valid number is entered
+        printf("Please enter a valid number:");
+        fgets(buffer, 3, stdin);  // Inputs the value from user
+
+        if (*buffer = '\n')  // No characters are entered
+            continue;
+        excess_len = 0;
+        while (getchar() != '\n') excess_len++;
```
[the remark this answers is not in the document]
niranjank2022: Yes, I did. If zero or negative integers of size less than that of the buffer entered, it is accepted. If their size is greater, then they are rejected. One case is if "00000000000" like this was entered, it is ignored rather accepting.

niranjank2022:
> > @CascadingCascade Can you please lend me hand in deep learning about the pointer concept in C? It is quite uncomfortable for me to solve any problems regarding pointers especially Double pointers and dynamic memory allocation. Can you please share any links?
> 
> Sure thing, I can help, but I really need to get some sleep first, it's 1am at my local time, I'll get back to you in the morning.

Sorry, 😅

CascadingCascade:
I searched the internet and didn't find anything satisfactory, so I figured I'll have to write a guide myself, here's the [link](https://github.com/CascadingCascade/Cassidys-guides/blob/main/Pointers%20in%20C.md). That should do it for the pointers, I'll write about dynamic allocation later

niranjank2022:
> I searched the internet and didn't find anything satisfactory, so I figured I'll have to write a guide myself, here's the [link](https://github.com/CascadingCascade/Cassidy-Guides/blob/main/Pointers%20in%20C.md). That should do it for the pointers, I'll write about dynamic allocation later Edit: I finish that second guide, but there are probably better resources out there🤔

Thanks a lot for your efforts. I would check this out for sure. But, if I have doubts, how can I ask you?

CascadingCascade:
> > I searched the internet and didn't find anything satisfactory, so I figured I'll have to write a guide myself, here's the [link](https://github.com/CascadingCascade/Cassidy-Guides/blob/main/Pointers%20in%20C.md). That should do it for the pointers, I'll write about dynamic allocation later Edit: I finish that second guide, but there are probably better resources out there🤔
> 
> Thanks a lot for your efforts. I would check this out for sure. But, if I have doubts, how can I ask you?

Open an issue at the guide repository? This is definitely out of the scope of this PR.

niranjank2022:
> > > I searched the internet and didn't find anything satisfactory, so I figured I'll have to write a guide myself, here's the [link](https://github.com/CascadingCascade/Cassidy-Guides/blob/main/Pointers%20in%20C.md). That should do it for the pointers, I'll write about dynamic allocation later Edit: I finish that second guide, but there are probably better resources out there🤔
> > 
> > 
> > Thanks a lot for your efforts. I would check this out for sure. But, if I have doubts, how can I ask you?
> 
> Open an issue at the guide repository? This is definitely out of the scope of this PR.

Ok sure, got it :)

alexpantyukhin (approved):

math/fibonacci.c
```diff
@@ -1,23 +1,114 @@
-#include <stdio.h>
+/**
+ * @file
+ * @brief Program to print the nth term of the Fibonacci series.
```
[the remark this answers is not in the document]
niranjank2022: Is this the [link](https://en.wikipedia.org/wiki/Fibonacci_sequence)?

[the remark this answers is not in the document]
Panquesito7: Sure, looks good. 👍

math/fibonacci.c
```diff
@@ -1,23 +1,115 @@
...
+ * @details 
+ * Fibonacci series generally starts from 0 and 1. Every next term in
+ * the series is equal to the sum of the two preceding terms.
+ * For further info: https://en.wikipedia.org/wiki/Fibonacci_sequence
```
Panquesito7, line 7:
```suggestion
 * For further info: https://en.wikipedia.org/wiki/Fibonacci_sequence
 *
 * @author [Luiz Carlos Aguiar C](https://github.com/IKuuhakuI)
 * @author [Niranjan](https://github.com/niranjank2022)
```

math/fibonacci.c
```diff
@@ -1,23 +1,118 @@
...
+            // No characters are entered
+            endPtr == buffer)
+        {
+            continue;
+        }
+
+        break;
+    }
+
+    printf("\nEntered digit: %d\n", num);
```
Panquesito7, line 83: We might want to add this here in case people wonder why it takes a bit long to compute.
It'd be great if the time could be added as well. What do you think?
```suggestion
    printf("\nEntered digit: %d (result might take a while)\n", num);
```

niranjank2022: Sure!

math/fibonacci.c
```diff
@@ -1,23 +1,124 @@
...
-    int number;
+    // Performing the test
+    test();
+    printf("Tests passed...\n");
 
-    // Asks for the number that is in n position in Fibonnacci sequence
-    printf("Number: ");
-    scanf("%d", &number);
+    // Getting n
+    printf(
```
Panquesito7, line 110: We should probably check if the number's above 48, and if so, do not let the Fibonacci function run.

niranjank2022: Yes. And also not permit other invalid inputs

math/fibonacci.c
```diff
@@ -1,23 +1,124 @@
...
+ * For further info: https://en.wikipedia.org/wiki/Fibonacci_sequence
+ *
+ * @author [Luiz Carlos Aguiar C](https://github.com/IKuuhakuI)
+ * @author [Niranjan](https://github.com/niranjank2022)
+ */
 
-// Fibonnacci function
-int fib(int number)
+#include <assert.h>  /// for assert()
+#include <errno.h>  /// for errno - to determine whether there is an using strtol()
```
CascadingCascade, line 14:
```suggestion
#include <errno.h>  /// for errno - to determine whether there is an error while using strtol()
```

math/fibonacci.c
```diff
@@ -1,23 +1,124 @@
...
+#include <stdio.h>   /// for input, output
+#include <stdlib.h>  /// for exit() - to exit the program
+#include <time.h>   /// to calculate time taken by fib()
+/**
+ * @brief Determines the nth Fibonacci term
+ * @param number - n in "nth term" and it can't be negative as well as zero
+ * @return nth term in unsigned type
+ * @warning
+ * Only till 47th and 48th fibonacci element can be stored in 'int' and
+ * 'unsigned' respectively (takes more than 20 seconds to print)
```
CascadingCascade, line 24:
```suggestion
 * Only till 47th and 48th fibonacci element can be stored in `int` and
 * `unsigned int` respectively (takes more than 20 seconds to print)
```

math/fibonacci.c
CascadingCascade: I thought the `getinput()` function already took care of that? Or the idea is to stop the program instead of re-request input? I must say I'm a bit confused about what the two of you are talking about.

CascadingCascade (approved):

math/fibonacci.c
Panquesito7: Yeah, the function already manages that (kind of missed it out).

Panquesito7 (dismissed):

alexpantyukhin (approved):

Panquesito7 (approved):

PR #1243, opened 2023-03-31
Fix: Segmentation fault in merge_sort.c and print message

PKGzGMTH:
#### Description of Change
* Fix: Segmentation fault in merge_sort.c
* Feat: Print message of input and output.

#### Checklist

Notes:
* Handle error if array can't malloc.
* Exit program if array size is not greater than 0.


sorting/merge_sort.c
```diff
@@ -33,6 +33,8 @@ void swap(int *a, int *b)
 void merge(int *a, int l, int r, int n)
 {
     int *b = (int *)malloc(n * sizeof(int)); /* dynamic memory must be freed */
+    if (!b) /* return to main if can't malloc memory */
+        return ;
```
CascadingCascade, line 37: Maybe you should just make it print a message then exit, it's not like it can resume normal operation afterwards anyway

CascadingCascade (commented):
TBH, I'm not exact sure why the original author chose to use dynamic storage allocation, if we are really dealing with arrays that big we wouldn't enter all the data by hand would we?

sorting/merge_sort.c
```diff
@@ -33,6 +33,11 @@ void swap(int *a, int *b)
 void merge(int *a, int l, int r, int n)
 {
     int *b = (int *)malloc(n * sizeof(int)); /* dynamic memory must be freed */
+    if (b == NULL) /* return to main if can't malloc memory */
+    {
+        printf("Can't Malloc! Please try again.");
+        exit(1);
```
CascadingCascade, line 39: I would suggest you to use the macro instead of a magic number, although it probably does not matter given how widely `1` is recognized to be an abnormal exit code
```suggestion
        exit(EXIT_FAILURE);
```

sorting/merge_sort.c
```diff
@@ -33,6 +33,11 @@ void swap(int *a, int *b)
 void merge(int *a, int l, int r, int n)
 {
     int *b = (int *)malloc(n * sizeof(int)); /* dynamic memory must be freed */
+    if (b == NULL) /* return to main if can't malloc memory */
```
CascadingCascade, line 36: Removes this outdated comment since the program exits now
```suggestion
    if (b == NULL)
```

sorting/merge_sort.c
```diff
@@ -101,18 +106,32 @@ void merge_sort(int *a, int n, int l, int r)
...
 {
     int *a, n, i;
+    printf("Enter Array size: ");
     scanf("%d", &n);
+    if (n <= 0) /* exit program if arraysize is not greater than 0 */
+    {
+        printf("Array size must be Greater than 0!\n");
+        return 1;
+    }
     a = (int *)malloc(n * sizeof(int));
```
CascadingCascade, line 116: Here's a little language lawyer comment: I've heard there's a difference between `n * sizeof(int)` and `sizeof(int) * n`, the first causes the compiler to do the multiplication in `int` type instead of `size_t`, so the result gets cast back and forth between `int` and `size_t`, but I doubt if this is true, modern compilers are probably not that stupid

CascadingCascade (dismissed):
Looks good enough to me

Panquesito7 (approved):
Looks good. Thanks! 🚀

alexpantyukhin (approved):
Looks good to me

PR #1252, opened 2023-04-27
feat: add LeetCode problem 434

PalAditya:
Adds solution of problem 434 for leetcode. Beats 100% Time, 97.18% space

#### Checklist

### DIFF
--- a/leetcode/src/434.c
+++ b/leetcode/src/434.c
@@ -1,0 +1,27 @@
+// Given a string s, returns the number of segments in the string.
+int countSegments(char *s) {
+    int i, l = strlen(s), segments = 0;
+    if (l == 0) { // Return early for empty string
+        return 0;
+    }
+
+    for (i = 0; i < l; i++) {
+        if (s[i] != ' ') {
+            break; // Skip to first non-space character
+        }
+    }
+
+    //Start from the first non-space character index, calculated above
+    for (; i < l;) {
+        while (i < l && s[i] != ' ') { // Iterate till current character is not a space
+           i++;
+        } 
+        //We hit a space, so increment count
+        segments++;
+        while (i < l && s[i] == ' ') { //Skip consecutive spaces
+            i++;
+        }
+    }
+
+    return segments;
+}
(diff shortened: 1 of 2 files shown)


PalAditya:
TimeSpace

Proof of it being optimal 😄

Panquesito7 (dismissed):
Thanks! 🚀

leetcode/src/434.c
```diff
@@ -0,0 +1,27 @@
+// Given a string s, returns the number of segments in the string.
+int countSegments(char *s) {
+    int i, l = strlen(s), segments = 0;
+    if (l == 0) { // Return early for empty string
```
alexpantyukhin, line 4: do we need this guard condition? I think if no, theresult will be still 0 in the end.

alexpantyukhin (commented):
What if to simplify the solution with only one loop? Something like that:

```c
int countSegments(char * s){
    int sLen = strlen(s);
    int prevSpace = 1;
    int result = 0;
    char currChar;

    for (int i = 0; i < sLen; i++){
        currChar = s[i];
        if (s[i] != ' ' && prevSpace){
            result++;
        }

        prevSpace = (currChar == ' ');
    }

    return result;
}
```

It beats 100% :)

leetcode/src/434.c
PalAditya: Yep, you're right

PalAditya:
@alexpantyukhin - Your version is simply superior, so I can't think of a good reason to not use it directly 😄 
I have added one explanatory comment which maybe unnecessary, but hopefully still explains the correctness of the logic as I see it.
P.S: Not sure if this PR is still valid, since I have pretty much copy-pasted your version apart from the comment. Please let me know

Panquesito7 (approved):
> Not sure if this PR is still valid, since I have pretty much copy-pasted your version apart from the comment. Please let me know

Both your contribution and @alexpantyukhin's contribution will be mentioned and kept, so I don't think there's much to worry about :) Thanks.

PR #1257, opened 2023-05-05
feat: add LeetCode problem 69

hemanth8819:
#### Checklist

- [x] Added description of change
- [x] Added file name matches [File name guidelines](https://github.com/TheAlgorithms/C/blob/master/CONTRIBUTING.md#File-Name-guidelines)
- [x] Added tests and example, test must pass
- [x] Relevant documentation/comments is changed or added
- [x] PR title follows semantic [commit guidelines](https://github.com/TheAlgorithms/C/blob/master/CONTRIBUTING.md#Commit-Guidelines)
- [x] Search previous suggestions before making a new one, as yours may be a duplicate.
- [x] I acknowledge that all my contributions will be made under the project's license.

Notes:

PR #1259, opened 2023-05-09
feat: add LeetCode problem 69

hemanth8819:
Here is the code for the problem 69 of leetcode as there are many ways to do it we programmers need to find the most optimal way to solve a problem statement so i used binary-search approach inorder to solve it. All  suggestions are accepted!!!

#### Checklist

Notes:

### DIFF
--- a/leetcode/src/69.c
+++ b/leetcode/src/69.c
@@ -5,9 +5,9 @@ int start=0;
         long long int ans=0;
         while(start <= end){
                long long int mid=(start+end)/2;
+                long long int val=mid*mid;
                if(mid*mid == x){
-                         ans=mid;
-                         break;
+                         return mid;
                 }
 //if mid is less than the square root of the number(x) store the value of mid in ans.
                if( mid*mid < x){
--- a/leetcode/src/69.c
+++ b/leetcode/src/69.c
@@ -1,0 +1,24 @@
+//using binary search method is one of the efficient one for this problem statement.
+int mySqrt(int x){
+int start=0;
+        int end=x;
+        long long int ans=0;
+        while(start <= end){
+               long long int mid=(start+end)/2;
+               if(mid*mid == x){
+                         ans=mid;
+                         break;
+                }
+//if mid is less than the square root of the number(x) store the value of mid in ans.
+               if( mid*mid < x){
+                         ans = mid;
+                       start = mid+1;
+               }
+//if mid is greater than the square root of the number(x) then ssign the value mid-1 to end.               
+               if( mid*mid > x){
+                       end = mid-1;
+               }
+        }
+        return ans;
+
+}
--- a/leetcode/src/69.c
+++ b/leetcode/src/69.c
@@ -1,17 +1,17 @@
-//using binary search method is one of the efficient one for this problem statement.
+//using the binary search method is one of the efficient ones for this problem statement.
 int mySqrt(int x){
 int start=0;
         int end=x;
         long long int ans=0;
         while(start <= end){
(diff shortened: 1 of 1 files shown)


alexpantyukhin:
Why you closed the previos MR? you could make changes in the context of the prev.

CascadingCascade:
> Why you closed the previos MR? you could make changes in the context of the prev.

Check out #1257 (which is closed by me for being a duplicate), they refuse to do that for whatever reason

leetcode/src/69.c
```diff
@@ -0,0 +1,24 @@
...
+                         ans=mid;
+                         break;
+                }
+//if mid is less than the square root of the number(x) store the value of mid in ans.
+               if( mid*mid < x){
+                         ans = mid;
+                       start = mid+1;
+               }
+//if mid is greater than the square root of the number(x) then ssign the value mid-1 to end.               
+               if( mid*mid > x){
```
alexpantyukhin, line 18: You are doing multiplication 3 times per loop in bad case. You can store the  result of `mid * mid` in variable and use it.

leetcode/src/69.c
```diff
@@ -0,0 +1,24 @@
+//using binary search method is one of the efficient one for this problem statement.
+int mySqrt(int x){
+int start=0;
+        int end=x;
+        long long int ans=0;
+        while(start <= end){
+               long long int mid=(start+end)/2;
+               if(mid*mid == x){
+                         ans=mid;
```
alexpantyukhin, line 9: May just `return mid;`

leetcode/src/69.c
hemanth8819: ok thanks for the suggestion i will consider that.

leetcode/src/69.c
hemanth8819: sorry I can't return mid because when mid*mid < number  we need to store the max value of mid so I used ans to store the value of the mid.But i will return mid for the case when mid*mid == number .
thank you for the suggestions!!!

leetcode/src/69.c
```diff
@@ -0,0 +1,24 @@
...
+        int end=x;
+        long long int ans=0;
+        while(start <= end){
+               long long int mid=(start+end)/2;
+                long long int val=mid*mid;
+               if(mid*mid == x){
+                         return mid;
+                }
+//if mid is less than the square root of the number(x) store the value of mid in ans.
+               if( mid*mid < x){
```
alexpantyukhin, line 13: You have the `val=mid*mid` but you still multiply it 3 times and check the value. You can use val instead.

alexpantyukhin (approved):

leetcode/src/69.c
```diff
@@ -0,0 +1,24 @@
+//using binary search method is one of the efficient one for this problem statement.
+int mySqrt(int x){
+int start=0;
+        int end=x;
+        long long int ans=0;
+        while(start <= end){
+               long long int mid=(start+end)/2;
+                long long int val=mid*mid;
```
Panquesito7, line 8: There seem to be a few indentation issues here (and in other parts of the code as well). Please fix those.

Panquesito7 (changes requested):

leetcode/src/69.c
hemanth8819: Yeah, thanks for figuring it out.I have made the respective changes for that.

alexpantyukhin (approved):

Panquesito7 (approved):
Thanks! 🚀

Issue #1269, opened 2023-06-10
[BUG] Segmentation fault in `queue.c`


Arash1381-y:
### Description

In the queue model (path : data_structures/queue/queue.c). dequeue won't work due to a simple bug on the last if clause of it. take a look:
https://github.com/TheAlgorithms/C/blob/b1a8da69a87cd45e9db0cc340730e4a5e3b36201/data_structures/queue/queue.c#L65

as you can see if we have only one variable in our queue on dequeue the HEAD will be null and assigning NULL to NULL->next will cause a segmentation fault.

### Expected behavior

not to cause fault =)

### Actual behavior

case segmentation fault

### Possible fix

simply put line 65 in its above clause

### Steps to reproduce

1. call create
2. enqueue anything you like 
3. dequeue it 

### Context

It just don't work =)

PR #1270, opened 2023-06-10
fix  bug in

Arash1381-y:
fix the bug i issued [here](https://github.com/TheAlgorithms/C/issues/1269)

### DIFF
--- a/data_structures/queue/queue.c
+++ b/data_structures/queue/queue.c
@@ -61,8 +61,10 @@ int deque()
         if (head->pre == NULL)
             head = NULL;
         else
+        {
             head = head->pre;
-        head->next = NULL;
+            head->next = NULL;
+        }
     }
     return returnData;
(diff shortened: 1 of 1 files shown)


data_structures/queue/queue.c
```diff
@@ -61,8 +61,10 @@ int deque()
         if (head->pre == NULL)
             head = NULL;
         else
+        {
             head = head->pre;
-        head->next = NULL;
+            head->next = NULL;
```
Panquesito7, line 66: There don't seem to be any changes rather than the brackets. 🤔
Could you please explain if you made any changes?

alexpantyukhin: hm.. I think there is no bug. The `next` element of the `pre` should be NULL in this case. (One thing left is to `free`, but it's  another).

PR #1298, opened 2023-08-30
Create average.c

lukasb1b:
adding an average function with a test case

#### Checklist

Notes:

### DIFF
--- a/.github/workflows/awesome_workflow.yml
+++ b/.github/workflows/awesome_workflow.yml
@@ -26,11 +26,6 @@ jobs:
         uses: TheAlgorithms/scripts/formatter@main
         with:
           filetypes: .c,.h
-      - name: Update DIRECTORY.md
-        run: |
-          wget https://raw.githubusercontent.com/TheAlgorithms/scripts/main/build_directory_md.py
-          python3 build_directory_md.py C . .c,.h leetcode/ > DIRECTORY.md
-          git commit -m "updating DIRECTORY.md" DIRECTORY.md || true
       - name: Get file changes
         run: |
           git branch
--- a/.github/workflows/directory_writer.yml
+++ b/.github/workflows/directory_writer.yml
@@ -1,0 +1,29 @@
+name: Directory writer
+on:
+  schedule:
+  #        ┌───────────── minute (0 - 59)
+  #        │  ┌───────────── hour (0 - 23)
+  #        │  │ ┌───────────── day of the month (1 - 31)
+  #        │  │ │ ┌───────────── month (1 - 12 or JAN-DEC)
+  #        │  │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
+  #        │  │ │ │ │
+  #        │  │ │ │ │
+  #        │  │ │ │ │
+  #        *  * * * *
+  - cron: '0 0 * * *'
+  workflow_dispatch:
+jobs:
+  build:
+    if: github.repository == 'TheAlgorithms/C' # We only need this to run in our repository.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Build directory
+        uses: TheAlgorithms/scripts/directory_md@main
+        with:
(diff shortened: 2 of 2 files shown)


math/average.c
```diff
@@ -0,0 +1,25 @@
...
+    
+    while(avg[size+1]!='\0'){
+        size++;
+    }
+
+    for (int i = 0; i < size; i++) {
+        result += avg[i];
+    }
+
+    return result / size;
```
alexpantyukhin, line 18: what if the size is zero?

math/average.c
```diff
@@ -0,0 +1,25 @@
+#include <stdio.h>
+#include <assert.h>
+
+//average of an array
+
+double average(int avg[]) {
+    double result = 0;
+    int size=0;
+    
+    while(avg[size+1]!='\0'){
```
alexpantyukhin, line 10: I think this case you can find the array size without a loop. For example: `sizeof(avg) / sizeof(int)` What do you think?

math/average.c
```diff
@@ -0,0 +1,25 @@
+#include <stdio.h>
+#include <assert.h>
+
+//average of an array
+
+double average(int avg[]) {
```
alexpantyukhin, line 6: why avg? is it the array of avgs? Maybe values? What do you think?

math/average.c
lukasb1b: lets do something like if (size== 0) {
        perror("empty array");
    }

math/average.c
lukasb1b: I tried it too but i did not get it to work

math/average.c
lukasb1b: can you explain this please

math/average.c
CascadingCascade: IIRC, `sizeof(avg) / sizeof(int)` won't work here, an array degrades into a pointer to its first element when passed into a function, and dividing the size of a pointer obviously does nothing

alexpantyukhin: yes, you are right.

math/average.c
rbevin777, line 6: Don't forget to add the function header. This should look something like this
```
/**
 * @brief    This is a brief description of what the function does
 * @param param_name - This is a brief description of the parameter
 * @return return_name - Biref description of the return value itself.
 */
```

math/average.c
rbevin777, line 18: It'd be useful to declare a variable like this so that if you or someone else needs to debug the code it is eaiser to trace.
```
int avg_val = result / size;
return avg_val;
```

math/average.c
```diff
@@ -0,0 +1,25 @@
...
+
+    for (int i = 0; i < size; i++) {
+        result += avg[i];
+    }
+
+    return result / size;
+}
+
+int main() {
+    int test[] = {1,2,3,4,5};
```
rbevin777, line 22: You could add a few different tests to ensure the robustness, so for example using all negative numbers, a mix of negative and positive, you could go for large numbers close to the maximum size of an `int` and see what happens there (most likely an overflow and the wrong result) which is what you want to see from tests anyway. Play about with testing. It's great at highlighting the faults.

rbevin777, line 22: Another comment in regards to testing, there is a common testing documentation format called [gherkin](https://cucumber.io/docs/gherkin/reference/) The basics of it are 

```
// Given we have an array fo values

// When we try to work out the average value of the array

// Then we should get the expected result
```

This is a simple example but simple is good.

rbevin777 (changes requested):
Good module, also don't be disheartened by the number of comments either! 😁

lukasb1b:
I changed all of your requests exept of the last one and the last test is not passing but i dont know why because it prints15.000000-15.000000 and I think that should be 0 any fixes

lukasb1b:
I made a workaround at the tests by setting the last value 0 but otherwise i don't think there is a solution to calulate the array length. I just can think of creating a parameter for the length

Adlinke:
I am so excited to having you! Thank you so much!

PR #1377, opened 2023-12-23
fix: add missing return in removeValueCArray

honey-yogurt:
#### removeValueCArray lacks a return statement in one of its branches.
```
int removeValueCArray(CArray *array, int position)
{
    if (position >= 0 && position < array->size)
    {
        if (array->array[position] != 0)
        {
            array->array[position] = 0;
            // here missing.
            return SUCCESS;
        }
        else
            return POSITION_EMPTY;
    }
    return INVALID_POSITION;
}
```


realstealthninja (approved):
Nicely spotted

PR #1463, opened 2025-03-19
fix: Prevent segmentation fault at Simple generic Stack

Artursrossi:
#### Description of Change
- Prevent segmentation fault error at stack.c grow function
  - After malloc, there is an iteration without any check if malloc returned null

#### Checklist

Notes: Used realloc instead of malloc

### DIFF
--- a/data_structures/stack/stack.c
+++ b/data_structures/stack/stack.c
@@ -45,17 +45,8 @@ void grow()
 {
     max += 10; /* increases the capacity */
 
-    int i;  // for the loop
-    void **tmp = malloc(sizeof(void *) * max);
-
-    /* copies the elements from the origin array in the new one. */
-    for (i = 0; i < max - 10; i++)
-    {
-        *(tmp + i) = *(array + i);
-    }
-    /*free the memory */
-    free(array);
-    array = tmp;
+    array = (void **)realloc(array, sizeof(void *) * max);
+    assert(array); /* tests whether pointer is assigned to memory. */
 }
 
 /* push: pushs the argument onto the stack */


data_structures/stack/stack.c
```diff
@@ -45,17 +45,8 @@ void grow()
...
-    /* copies the elements from the origin array in the new one. */
-    for (i = 0; i < max - 10; i++)
-    {
-        *(tmp + i) = *(array + i);
-    }
-    /*free the memory */
-    free(array);
-    array = tmp;
+    array = (void **)realloc(array, sizeof(void *) * max);
+    assert(array); /* tests whether pointer is assigned to memory. */
```
vil02, line 49: If asserts are disabled and `realloc` would return `NULL`, there will be a memory leak.

[C++ reference on `realloc`](https://en.cppreference.com/w/c/memory/realloc) states
> If there is not enough memory, the old memory block is not freed and null pointer is returned.

vil02 (changes requested):
In principle this is a change in a very good direction: standard functions are always better than manual work.

data_structures/stack/stack.c
Artursrossi: @vil02 I used assert() function just to maintain the file code standard. In other functions like initStack(), push() and pop(), the original author chose to use assert().

Even so, do you think this approach is better?:
if (array == NULL) {
    perror("Unable to allocate memory for stack array.");
    exit(1);
  }

Artursrossi (commented):
I used assert() function just to maintain the file code standard. In other functions like initStack(), push() and pop(), the original author chose to use assert().

data_structures/stack/stack.c
vil02: There is no problem with using assert. I simply said that the suggested code (although much better than the original one) can lead to a very tedious bug (it will not occur often, but when it will occur it will be hard to find).

In any case, the idea of explicitly checking with `array == NULL` looks reasonable.

PR #1516, opened 2025-10-26
feat(searching): add exponential_search2 with self-tests

hobostay:
#- What: Add searching/exponential_search2.c as an alternative exponential search implementation with Doxygen header and assert-based self-tests.
- Why: Complement existing exponential_search.c with clear range expansion and bounded binary search; verifies edge cases.
- Tests: Includes assert tests for present/absent values, boundaries, and empty input.
- Build: Picked up by searching/CMakeLists.txt via glob; C11 compliant; no external deps.


hobostay:
- Complexity: Time O(log n) via exponential range expansion + binary search; Space O(1) .
- Edge cases: empty array, out-of-range targets, boundaries covered by asserts.

PR #1558, opened 2026-02-19
fix: add missing return statements in doubly_linked_list.c

### LINKED ISSUE #1554: [BUG] Return statement in data_structures/linked_list/doubly_linked_list.c
### Description

There is a bug in the `doubly_linked_list.c` because some functions don't have a return statement for every path. The consequence is an undefined behaviour in some context. 

### Expected behavior

The result of the search function should be 1 if the value exists and 0 if not.

### Actual behavior

Undefined behaviour for some configuration.

### Possible fix

Add return statement to search, insert and delete functions.

### Steps to reproduce

# with gcc 13.3.0
gcc -o test_gcc doubly_linked_list.c
./test_gcc

3.000000        0.000000        20.000000       10.000000       5.000000
1
5.000000
0

# with clang 18.1.3
clang -o test_clang doubly_linked_list.c
./test_clang

3.000000        0.000000        20.000000       10.000000       5.000000
24027
5.000000
24027

### Context

For a personal project I needed to compile some C code with no dependencies. So I chose random programs from this repo. 

I noticed a bug with one of the algorithms, especially the doubly_linked_list.c under: data_structures/linked_list
The problem is that some functions in this program don't have a return statement for every path. 

For the functions `insert` and `delete` this is not crucial for the execution, but it is safer to add a `return list;` at the end. 

For the `search` function this is more problematic as the behaviour is undefined. For example working on my Ubuntu machine with GCC 13.3.0 and Clang 18.1.3 we have different results with the test present in the main function.

### Additional information

You have the warning for the non-return statement for every path with clang. You can enable them for gcc using the option `-Wall`.

There is only a minor change to make to fix the issue. I can propose the new version if needed with a small pull request. 
I didn't check the other algorithm under the linked_lidt/ folder as I was not working with them. But I would be happy to do it if there is a need.

danielalanbates:
#### Description of Change

Fixes three missing return statements in `data_structures/linked_list/doubly_linked_list.c` that caused undefined behavior:

1. **`search()`** (line 248): The recursive call `search(list->next, value)` was not preceded by `return`, so the function only correctly searched the head node. All deeper matches returned garbage.
2. **`insert()`**: When `pos <= 0`, execution fell off the end of the function without returning, which is undefined behavior for a non-void function.
3. **`delete()`**: Same issue as `insert()` — no return statement when `pos <= 0`.

#### References

Fixes #1554

#### Checklist

Notes: Minimal fix — only adds the missing `return` keywords, no other changes.


danielalanbates:
Thank you for your time reviewing this. We're withdrawing this PR. Apologies for any inconvenience.

